PR #39, opened 2015-03-06
standardize filters

snopoke:
All resources support the same filtering mechanism via URL parameters.

This will allow specifying other filters for the form endpoint without relying on the `_search` backdoor.


millerdev:
Can't really tell if this removes the `_search` backdoor, but I assume that if it does that is no longer used by anything.

PR #41, opened 2015-06-29
Add type support

snopoke:
@millerdev 

This allows you to map value to dates, timestamps, booleans.

Previously those would all be saved to SQL as strings.

Added docs here: https://confluence.dimagi.com/display/commcarepublic/CommCare+Data+Export+Tool#CommCareDataExportTool-E.TypeSupport

### DIFF
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -222,7 +222,7 @@ class JsonPathEnv(Env):
 def str2bool(val):
     if isinstance(val, bool):
         return val
-    return val and val.lower() in {'true', 't', '1'}
+    return val and str(val).lower() in {'true', 't', '1'}
 
 @unwrap
 def str2num(val):
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -1,10 +1,12 @@
 from __future__ import unicode_literals, print_function, absolute_import, division, generators, nested_scopes
+import functools
 import operator
 import six
 from itertools import chain
 
 from jsonpath_rw import jsonpath
 from jsonpath_rw.parser import parse as parse_jsonpath
+from commcare_export.misc import unwrap
 
 from commcare_export.repeatable_iterator import RepeatableIterator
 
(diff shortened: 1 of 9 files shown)


commcare_export/env.py
```diff
@@ -1,10 +1,12 @@
 from __future__ import unicode_literals, print_function, absolute_import, division, generators, nested_scopes
+import functools
```
millerdev: unused import

commcare_export/writers.py
```diff
@@ -141,7 +156,7 @@ def write_table(self, table):
         # Ensures the table is iterable; probably better to create a custom JSON handler that runs in constant space
         self.tables.append(dict(name=table['name'],
                                 headings=list(table['headings']),
-                                rows=[list(row) for row in table['rows']]))
+                                rows=[list([to_jvalue(v) for v in row]) for row in table['rows']]))
```
millerdev: `list()` call is unnecessary

commcare_export/writers.py
```diff
@@ -225,15 +253,9 @@ def compatible(self, source_type, dest_type):
...
-
-        # FIXME: Add datetime and friends
-        if isinstance(source_type, self.sqlalchemy.Integer):
-            # Integers can be cast to varch
-            return True
-
         if isinstance(source_type, self.sqlalchemy.String):
             if not isinstance(dest_type, self.sqlalchemy.String):
-                False
+                return False
```
millerdev: d'oh!

commcare_export/env.py
```diff
@@ -217,6 +219,36 @@ def emitted_tables(self):
 # Actual concrete environments, basically with built-in functions.
 #
 
+@unwrap
+def str2bool(val):
+    if isinstance(val, bool):
+        return val
+    return val and val.lower() in {'true', 't', '1'}
```
millerdev: Looks like this function may be called with non-string arguments, otherwise `isinstance` call above would be unnecessary. Is it safe to assume `val` will always be either a string or bool? If not, should this to use `str(val).lower()` (or something more fancy if `val` could contain non-ascii Unicode characters)?

commcare_export/env.py
```diff
@@ -217,6 +219,36 @@ def emitted_tables(self):
...
+
+@unwrap
+def str2num(val):
+    if val is None:
+        return None
+
+    try:
+        return int(val)
+    except ValueError:
```
millerdev: Would it make sense to also catch `TypeError`, in case `val` is not a string?

snopoke: I'd prefer to let this error

snopoke:
@millerdev thanks for the review, I've addressed your comments.

millerdev:
GTM when test pass

PR #46, opened 2016-11-02
implement multi-select functions

snopoke:
@gcapalbo


snopoke:
:white_check_mark:

commcare_export/env.py
```diff
@@ -250,11 +250,44 @@ def str2date(val):
...
+    if not val:
+        return None
+
+    try:
+        index = int(index)
+    except ValueError:
+        return "Error: index must be an integer"
+
+    try:
+        return str(val).split(' ')[index]
```
gcapalbo: might be better to use .split() rather than split(' ') since .split() is more forgiving of contiguous whitespace, or leading or trailing whitespace. or does that not matter here because we always expect to be passed a multi-select value?

snopoke: nice, updated: 18ae3e5

commcare_export/env.py
```diff
@@ -250,11 +250,44 @@ def str2date(val):
...
+    parts = str(val).split()
+    return reference in parts
+
+
+@unwrap('val')
+def count_selected(val):
+    if not val:
+        return None
+
+    return len(str(val).split(' '))
```
gcapalbo: i think there's just one more split() to update

snopoke:
:white_check_mark:

PR #49, opened 2016-11-28
--output docs

benrudolph:
@snopoke this seem ok?


snopoke:
Thanks, I've also added this info to the confluence page

PR #51, opened 2017-05-23
correctly handle unicode in map functions

snopoke:
https://manage.dimagi.com/default.asp?254676

### DIFF
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -254,7 +254,6 @@ def str2date(val):
     if not isinstance(val, six.string_types):
         val = str(val)
 
-    val = val.encode('utf8')
     try:
         date = parser.parse(val)
     except ValueError:
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -222,9 +222,17 @@ class JsonPathEnv(Env):
 
 @unwrap('val')
 def str2bool(val):
+    if val is None:
+        return False
+
     if isinstance(val, bool):
         return val
-    return val and str(val).lower() in {'true', 't', '1'}
+
+    if not isinstance(val, six.string_types):
+        val = str(val)
+
+    return val and val.lower() in {'true', 't', '1'}
+
 
 @unwrap('val')
 def str2num(val):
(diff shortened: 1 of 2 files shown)


commcare_export/env.py
```diff
@@ -242,7 +250,16 @@ def str2date(val):
     import dateutil.parser as parser
     if not val:
         return None
-    date = parser.parse(val)
+
+    if not isinstance(val, six.string_types):
+        val = str(val)
+
+    val = val.encode('utf8')
```
kaapstorm: Are you sure `val` can't already be a UTF-8 bytestring? If it is, this will throw an exception (UnicodeDecodeError in Py2 and AttributeError in Py3).

If you know `val` will be Unicode or ASCII then this is fine. But if it could be anything, you should probably check before calling `encode()`, maybe like
```
if not isinstance(val, bytes):  # or if isinstance(val, six.text_type):
    val = val.encode('utf8')
```

kaapstorm:
Looks good. Just the one question.

commcare_export/env.py
snopoke: Thanks, don't actually need the encoding at all.

snopoke:
Thanks for the feedback @millerdev. Looks like I need to brush up on my Python3....

commcare_export/env.py
```diff
@@ -222,9 +222,17 @@ def emitted_tables(self):
...
 def str2bool(val):
+    if val is None:
+        return False
+
     if isinstance(val, bool):
         return val
-    return val and str(val).lower() in {'true', 't', '1'}
+
+    if not isinstance(val, six.string_types):
+        val = str(val)
```
millerdev: Thinking a bit more about "We should probably use this everywhere..." In the context of `str2bool`, `val.decode('ascii', 'replace')` will work just fine, but in other contexts we may want to decode bytes using some other method, possibly `val.decode('utf-8', 'replace')`?

millerdev: I guess it somewhat depends on whether we want to always coerce to the desired type, even if it means losing some information, or if we want to raise errors for invalid input. If the former, then `val.decode('utf-8', 'replace')` should always be safe. If the latter, then maybe `val.decode('utf-8')` is better? Choosing which one probably depends on context.

snopoke: Since the data is coming from HQ API's which should always be `utf-8` encoded I'm going to go with `val.decode('utf-8')`.

snopoke: https://github.com/dimagi/commcare-export/pull/52

PR #56, opened 2017-08-04
re-enable python3 tests

snopoke:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
 python:
  - "2.7"
-# - "3.3"
+ - "3.3"
```
emord: Should probably either change this to 3.6, or add more languages (fwiw django 2.0 will only support 3.5+ so 3.5 & 3.6 seem like reasonable choices)

snopoke: updated

commcare_export/misc.py
```diff
@@ -41,7 +41,7 @@ def unwrap_val(val):
         if len(val) == 1:
             val = val[0]
         else:
-            val = map(unwrap_val, val)
+            val = list(map(unwrap_val, val))
```
millerdev: Nitpick: would be better as a list comprehension `[unwrap_val(v) for v in val]` as to only iterate over the items once (in Python 2)

snopoke: 592d667

PR #58, opened 2017-08-11
refactor checkpoint management

snopoke:


commcare_export/checkpoint.py
```diff
@@ -0,0 +1,85 @@
...
+           WHERE final = :final
+           AND query_file_md5 = :md5
+       """.format(self.table_name)
+        self.connection.execute(
+            self.sqlalchemy.sql.text(sql),
+            final=False,
+            md5=query_md5,
+        )
+
+    def _migrate_checkpoint_table(self, columns):
```
snopoke: This is much more strict than the previous 'migration' which was a 'best effort' migration.

tests/test_writers.py
```diff
@@ -316,3 +320,92 @@ def test_mysql_types(self):
...
+    """For each '_test_*' method on the class create a test case for each sql engine"""
+    engines = ('mysql_engine', 'postgres_engine')
+    methods = inspect.getmembers(cls, predicate=inspect.ismethod)
+    for name, method in methods:
+        if name.startswith('_test_'):
+            def _make_test_cases(test_method):
+                # closure to make sure test_method is the right method
+                for engine in engines:
+                    def test(self):
+                        with getattr(self, engine).connect() as conn:
```
millerdev: This will not work as expected. By the time the method gets called, the name `engine` will reference the last item in the `engines` tuple. In other words, `engine` will be `'postgres_engine'` for all tests generated by this `for` loop.

One way to work around this is to use a function to create the test method. Something like this:
```py
def make_test(engine):
    def test(self):
        with getattr(self, engine)...
            ...
    return test
```
Then in the loop you can do `test = make_test(engine)`. [Explanation](https://eev.ee/blog/2011/04/24/gotcha-python-scoping-closures/).

snopoke:
Thanks. I already have a closure for creating the test case so I'll just
add `engine` as an argument.

tests/test_writers.py
snopoke: a359e8f

PR #61, opened 2017-09-18
filter forms by server_modifed_on or received_on

snopoke:
Rather than requiring a reindex of all our form indices use a filter that works with new & old docs


snopoke:
:white_check_mark:

PR #66, opened 2017-12-01
Update regex check to allow use of ()s within the arg

mkangia:
[Ticket](https://manage.dimagi.com/default.asp?258796)

I could not run and test this because i could not finish set up but looks like the fix for this.
@snopoke do you have this running to quickly test it?


snopoke:
To test just install the requirements (`pip install .` and `pip install pytest`) the run `py.test tests/test_minilinq.py`

mkangia:
Thank You @snopoke. Added tests and made a correction to regex.

PR #67, opened 2017-12-08
set batch checkpoint date to correct date

snopoke:
The current code naively sets the checkpoint to the
datetime that the batch was downloaded as opposed to the
'since date' in the last item of the batch.

This means that any interruptions in the export
would not correctly restart from where they left off
but would instead only export data that was updated
since the last run.


snopoke:
This is a shocking bug :beetle:

millerdev:
Some tests would be nice.

PR #71, opened 2018-04-10
Implement api key authentication

pr33thi:
Ticket: https://manage.dimagi.com/default.asp?273839
Enabling authentication with username, password, and apikey, which allows the tool to be used when 2 factor auth is required for users in a domain.

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -45,8 +45,7 @@ def main(argv):
     parser.add_argument('--api-version', default=LATEST_KNOWN_VERSION)
     parser.add_argument('--project')
     parser.add_argument('--username')
-    parser.add_argument('--password')
-    parser.add_argument('--apikey')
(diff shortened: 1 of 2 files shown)


commcare_export/cli.py
```diff
@@ -46,7 +46,9 @@ def main(argv):
     parser.add_argument('--project')
     parser.add_argument('--username')
     parser.add_argument('--password')
-    parser.add_argument('--auth-mode', default='digest', help='Use "session" based auth or "digest" auth.')
+    parser.add_argument('--apikey')
```
snopoke: any reason not to just use the `--password` arg instead of adding a new one?

pr33thi: Changed it to us the `--password` arg, and added some help text here:
https://github.com/dimagi/commcare-export/pull/71/commits/f3ce37697c59423070378ee40bd0e9392ce72f1d

PR #74, opened 2018-05-04
switch to using alembic for migrations

snopoke:
This is so that we can support SQL Server.


migrations/versions/29c27e7e2bf6_rename_time_of_run_to_since_param.py
```diff
@@ -0,0 +1,29 @@
...
+
+Revision ID: 29c27e7e2bf6
+Revises: d3ce9dc9907a
+Create Date: 2018-05-03 12:01:13.046003
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+revision = '29c27e7e2bf6'
```
pr33thi: Where did these id's get generated?

Issue #59, opened 2017-08-20
Proper handling of empty fields to SQL


yozgoesdigital:
I have observed the Commcare Export tool has problems handling empty fields. 
System
CommCare Export tool version 0.14 (but also with the previous version)
Writing the data to Postgresql 9.6

When using the str2num option in the Map Via column a traceback is given with the error: 
`ValueError: could not convert string to float:`
It appears that it happens with all not answered questions.
***\*\*update: it appears empty questions are written as empty string values\*\**** 

What I have tried:
When making a new table and changing the data type (e.g. by adding ::float8) I receive a similar error: `ERROR:  invalid input syntax for type double precision: ""`
When filtering non-empy results (e.g. by limiting the date range) the function works as expected.

Does somebody else experience similar problems?

PR #77, opened 2018-05-14
handle blank str2num

sravfeyn:
https://manage.dimagi.com/default.asp?276256
https://github.com/dimagi/commcare-export/issues/59

### DIFF
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -255,6 +255,8 @@ def str2num(val):
         return int(val)
     except ValueError:
         return float(val)
+    except ValueError:
+        return None
 
 
 @unwrap('val')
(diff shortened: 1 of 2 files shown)


commcare_export/env.py
```diff
@@ -255,6 +255,8 @@ def str2num(val):
         return int(val)
     except ValueError:
         return float(val)
+    except ValueError:
+        return None
```
snopoke: This should be
```python
try:
    return int(val)
except ValueError:
    try:
        return float(val)
    except ValueError:
        return None
```

sravfeyn:
Oops, thanks!

snopoke:
rebased off mater to fix build

PR #84, opened 2018-06-29
set checkpoint manager when cloning

snopoke:
Looks like checkpointing has been broken since: https://github.com/dimagi/commcare-export/pull/83

Affects v 0.22.0

### DIFF
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -83,7 +83,19 @@ class CommCareHqClient(object):
         else:
             raise Exception('Unknown auth mode: %s' % mode)
 
-        return CommCareHqClient(url=self.url, project=self.project, version=self.version, session=session, auth=auth)
+        return self._clone_with(session=session, auth=auth)
+
+    def _clone_with(self, **kwargs):
+        kwargs.setdefault('version', self.version)
+        kwargs.setdefault('session', self.__session)
+        kwargs.setdefault('auth', self.__auth)
+        cloned = CommCareHqClient(
+            self.url,
+            self.project,
+            **kwargs
+        )
+        cloned.set_checkpoint_manager(self._checkpoint_manager, **self._checkpoint_kwargs)
+        return cloned
 
     def get(self, resource, params=None):
         """
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -83,16 +83,14 @@ class CommCareHqClient(object):
         else:
             raise Exception('Unknown auth mode: %s' % mode)
 
-        return self._clone_with(session=session, auth=auth)
+        return self._clone_with(session, auth)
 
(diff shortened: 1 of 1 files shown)


snopoke:
This highlights the need for some tests of the cli code.

commcare_export/commcare_hq_client.py
```diff
@@ -83,7 +83,19 @@ def authenticated(self, username=None, password=None, mode=AUTH_MODE_SESSION):
...
+        kwargs.setdefault('version', self.version)
+        kwargs.setdefault('session', self.__session)
+        kwargs.setdefault('auth', self.__auth)
+        cloned = CommCareHqClient(
+            self.url,
+            self.project,
+            **kwargs
+        )
+        cloned.set_checkpoint_manager(self._checkpoint_manager, **self._checkpoint_kwargs)
+        return cloned
```
esoergel: I'm new to this codebase, can you walk me through what's going on?  How was it broken before?  What does "clone" mean in this context - get a new client object that's a copy of the last one?

commcare_export/commcare_hq_client.py
```diff
@@ -83,7 +83,19 @@ def authenticated(self, username=None, password=None, mode=AUTH_MODE_SESSION):
         else:
             raise Exception('Unknown auth mode: %s' % mode)
 
-        return CommCareHqClient(url=self.url, project=self.project, version=self.version, session=session, auth=auth)
+        return self._clone_with(session=session, auth=auth)
+
+    def _clone_with(self, **kwargs):
+        kwargs.setdefault('version', self.version)
+        kwargs.setdefault('session', self.__session)
+        kwargs.setdefault('auth', self.__auth)
```
esoergel: It looks like `_clone_with` is called in one place, and `session` and `auth` are both provided.  Do you expect this to be called elsewhere where you may want the default values?

commcare_export/commcare_hq_client.py
pr33thi: ^I could use this too!

snopoke: Sure. I'm not sure this is really necessary and may just be a holdover from the original design. As the docstring for this method says: 

    Returns a freshly authenticated CommCareHqClient with a new session.
    This is safe to call many times and each of the resulting clients
    remain independent, so you can log in with zero, one, or many users.

I don't think that's a real requirement since we only ever auth with a single user per process. There are one or two other uses of this class [outside the library](https://github.com/search?q=CommCareHqClient&type=Code) but probably nothing to bear in mind in terms of backwards compatabilty.

However since this is going to be released as a bug fix on an existing release I don't want to make a larger change.

commcare_export/commcare_hq_client.py
snopoke: No I guess not: 59df155

esoergel:
:+1:

PR #85, opened 2018-07-02
optimize query to prevent pulling data sources repeatedly

snopoke:
Currently each table output will do it's own API query to get the data. This PR adds functionality to allow outputting multiple tables from the same API query.

This is useful if you are doing any of the following;
* form + repeats
* form + case references
* case + case history
* case + indexes

Can review by commit :fish:


commcare_export/writers.py
```diff
@@ -146,19 +146,29 @@ def write_table(self, table):
...
         self.book.save(self.file)
 
+
 class JValueTableWriter(TableWriter):
     """
     Write tables to JSON-friendly in-memory values
     """
 
     def __init__(self):
-        self.tables = []
```
pr33thi: How come you changed it from a list to a dict? I guess this gets accessed by name somewhere?

snopoke: So previously each table would only get 'written' once. With the change to apply multiple `Emit` expressions for a single data source `write_table` function will get called once for every document so we need to be able to update existing tables.

commcare_export/excel_query.py
```diff
@@ -180,16 +180,18 @@ def compile_source(worksheet):
...
                        body=Reference(str(data_source_jsonpath)))
 
-def compile_sheet(worksheet, mappings=None, missing_value=None):
+def compile_sheet(worksheet, mappings=None, missing_value=None, ignore_empty=True):
     mappings = mappings or {}
     source_expr = compile_source(worksheet)
 
     output_table_name = worksheet.title
-    output_headings = get_column_by_name(worksheet, 'field') # It is unfortunate that this is duplicated here and in `compile_fields`
+    output_headings = get_column_by_name(worksheet, 'field')  # It is unfortunate that this is duplicated here and in `compile_fields`
```
pr33thi: This comment might be outdated- I don't see a similar line in `compile_fields`

commcare_export/excel_query.py
```diff
@@ -180,16 +180,18 @@ def compile_source(worksheet):
...
     output_fields = compile_fields(worksheet, mappings=mappings)
 
     if not output_fields:
-        headings = headings = []
+        if ignore_empty:
```
pr33thi: I'm having trouble understanding this variable name- to me ignore_empty implies that not having output fields _wouldn't_ raise an exception?

commcare_export/excel_query.py
snopoke: The comment is referring to the same call on line [103](https://github.com/dimagi/commcare-export/pull/85/files#diff-a06a76e2710b6183b752bc4e0cfacb87R103):
```
fields = without_empty_tail(get_column_by_name(worksheet, 'field') or [])
```

Probably not a very useful comment though. I'll remove it.

snopoke: 2b1854d

commcare_export/excel_query.py
```diff
@@ -229,19 +235,44 @@ def compile_workbook(workbook, missing_value=None, ignore_empty=True):
...
+            missing_value=missing_value
+        )
+        for sheet in parsed_sheets
+    ]
 
+
+def get_queries_from_excel(workbook, missing_value=None):
+    parsed_sheets = parse_workbook(workbook)
+    queries = compile_queries(parsed_sheets, missing_value)
+    return List(queries) if len(queries) > 1 else queries[0]
```
esoergel: This is great - big fan of splitting up the parsing and the processing stages.

commcare_export/cli.py
```diff
@@ -147,15 +148,20 @@ def main_with_args(args):
...
+        try:
+            query = _get_query(
+                args.query,
+                args.missing_value,
+                writer.supports_multi_table_write,
+                writer.max_column_length,
+            )
+        except LongFieldsException as e:
+            print(e.message)
+            return 1
```
esoergel: nit: since this is no longer `exit(1)`, it's not obvious what `return 1` means - could make a constant like `EXIT_STATUS_ERROR` or something.

snopoke: 961499b

PR #86, opened 2018-07-03
refactor and simplify

snopoke:
The main goal with this is to reduce the complexity of the cli code
This is a pure refactor and shouldn't change any actual functionality.

Can review per commit :blowfish:

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -102,7 +102,26 @@ def main(argv):
             stats.print_stats(100)
 
 
-def _get_query(query_arg, missing_value, combine_emits, max_column_length):
+def _get_query(args, writer):
+    # Reads as excel if it is a file name that looks like excel, otherwise reads as JSON,
+    # falling back to parsing arg directly as JSON, and finally parsing stdin as JSON
+    if args.query:
+        return _get_excel_query(
+            args.query,
+            args.missing_value,
+            writer.supports_multi_table_write,
+            writer.max_column_length,
+        )
+    else:
+        try:
+            return MiniLinq.from_jvalue(json.loads(sys.stdin.read()))
+        except Exception as e:
+            raise Exception(
+                "Failure reading query from console input. "
+                "Try using the '--query' parameter to pass your query as an Excel file", e
+            )
+
+def _get_excel_query(query_arg, missing_value, combine_emits, max_column_length):
     if os.path.exists(query_arg):
         if os.path.splitext(query_arg)[1] in ['.xls', '.xlsx']:
             import openpyxl
@@ -141,32 +160,16 @@ def main_with_args(args):
     run_start = datetime.utcnow()
 
     writer = _get_writer(args.output_format, args.output, args.strict_types)
-    
-    # Reads as excel if it is a file name that looks like excel, otherwise reads as JSON, 
-    # falling back to parsing arg directly as JSON, and finally parsing stdin as JSON
-    if args.query:
-        try:
-            query = _get_query(
-                args.query,
-                args.missing_value,
-                writer.supports_multi_table_write,
-                writer.max_column_length,
-            )
-        except LongFieldsException as e:
-            print(e.message)
-            return EXIT_STATUS_ERROR
 
-        if not query:
-            print('Query file not found: %s' % args.query)
-            return EXIT_STATUS_ERROR
-    else:
-        try:
-            query = MiniLinq.from_jvalue(json.loads(sys.stdin.read()))
-        except Exception as e:
-            raise Exception(
-                "Failure reading query from console input. "
-                "Try using the '--query' parameter to pass your query as an Excel file", e
-            )
+    try:
+        query = _get_query(args, writer)
+    except LongFieldsException as e:
+        print(e.message)
+        return EXIT_STATUS_ERROR
+
+    if not query:
+        print('Query file not found: %s' % args.query)
+        return EXIT_STATUS_ERROR
 
     if args.dump_query:
         print(json.dumps(query.to_jvalue(), indent=4))
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -155,7 +155,7 @@ def _get_writer(output_format, output, strict_types):
         raise Exception("Unknown output format: {}".format(output_format))
 
 
-def get_since_until(args, checkpoint_manager):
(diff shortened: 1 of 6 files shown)


kaapstorm:
I really like how you've simplified all of this.

I have a question about one aspect: I agree that `CommcareHqClient._clone_with()` and `CommcareHqClient.authenticated()` feel awkward, and that the `CommcareHqClientBuilder` factory class is a more explicit and familiar way to do the same thing.

And I like that the `_get_api_client()` function simplifies calling `CommcareHqClientBuilder`. 

Now that you have the function, can you use it to achieve the purpose of the factory class, and drop `CommcareHqClientBuilder`? Maybe something like ...
```python
def _get_api_client(args, checkpoint_manager, commcarehq_base_url):
    auth = CommCareHqClient.get_auth(args.username, args.password, args.auth_mode)
    return CommCareHqClient(
       url=commcarehq_base_url,
       project=args.project,
       version=args.api_version,
       auth=auth,
       checkpoint_manager=checkpoint_manager
   )
```
... where `CommCareHqClient.get_auth()` is a static method to return a `HTTPDigestAuth` or `ApiKeyAuth` instance?

The benefit is that you could replace a class with a method. The disadvantage is that you wouldn't be able to check whether `session` will be set in advance, but `CommCareHqClient.__init__()` would still check that.

This could be completely off the mark though. I'm not familiar with how the CLI gets used, and I could be missing something important.

snopoke:
@kaapstorm that is an option since the `CommCareHqClient` class is only intended for use within this project. What I do like about the builder is that it gives a clearer API for *how* it should be used.

The other option, which may actually be better, is to remove the ability to supply a `session` (which is only used by tests) and just add the auth args to the constructor.
```python
    def __init__(self, url, project, username, password, auth_mode=DIGEST, version=LATEST_KNOWN_VERSION, checkpoint_manager=None):
```

commcare_export/commcare_hq_client.py
```diff
@@ -148,20 +147,16 @@ def iterate_resource(resource=resource, params=params):
                 
         return RepeatableIterator(iterate_resource)
 
-    def set_checkpoint_manager(self, manager, **checkpoint_kwargs):
+    def set_checkpoint_manager(self, manager):
```
esoergel: :+1:

commcare_export/cli.py
```diff
@@ -191,12 +191,12 @@ def main_with_args(args):
...
-                                  version=args.api_version)
-
-    if checkpoint_manager:
-        api_client.set_checkpoint_manager(checkpoint_manager)
+    api_client = CommCareHqClient(
+        url=commcarehq_base_url,
+        project=args.project,
+        version=args.api_version,
+        checkpoint_manager=checkpoint_manager
+    )
```
esoergel: Nice - big fan of passing all dependencies to the constructor

commcare_export/cli.py
```diff
@@ -155,6 +155,21 @@ def _get_writer(output_format, output, strict_types):
         raise Exception("Unknown output format: {}".format(output_format))
 
 
+def get_since_until(args, checkpoint_manager):
```
pr33thi: Unclear function name, maybe rename or add a comment

commcare_export/cli.py
```diff
@@ -49,8 +44,8 @@ def main(argv):
     parser.add_argument('--project')
     parser.add_argument('--username')
     parser.add_argument('--password', help='Enter password, or if using apikey auth-mode, enter the api key.')
-    parser.add_argument('--auth-mode', default='digest', help='Use "session" based auth, "digest" auth, or'
-                                                              ' "apikey" auth (for two factor enabled domains).')
+    parser.add_argument('--auth-mode', default='digest', choices=['digest', 'apikey'],
+                        help='Use "digest" auth, or "apikey" auth (for two factor enabled domains).')
```
esoergel: What's the impetus behind this change?  Is it the case that session auth doesn't offer any advantage over digest in this context and so there's no point to support both?

commcare_export/cli.py
```diff
@@ -155,6 +155,21 @@ def _get_writer(output_format, output, strict_types):
...
+    if not args.since and not args.start_over and os.path.exists(args.query) and checkpoint_manager:
+        with checkpoint_manager:
+            args.since = checkpoint_manager.get_time_of_last_run()
+
+        if args.since:
+            logger.debug('Last successful run was %s', args.since)
+        else:
+            logger.warn('No successful runs found, and --since not specified: will import ALL data')
+
+    since = dateutil.parser.parse(args.since) if args.since else None
```
pr33thi: Same comment as above! I think a more descriptive var name name for `since` and `until` would help me follow this

commcare_export/cli.py
```diff
@@ -203,9 +203,12 @@ def main_with_args(args):
 
     checkpoint_manager = None
     if writer.support_checkpoints:
-        checkpoint_manager = CheckpointManager(args.output, args.query, misc.digest_file(args.query))
-        with checkpoint_manager:
-            checkpoint_manager.create_checkpoint_table()
+        if not os.path.exists(args.query):
```
pr33thi: If it's possible the path doesn't exist, is it possible it's empty?

commcare_export/commcare_hq_client.py
```diff
@@ -26,6 +26,37 @@
 
 LATEST_KNOWN_VERSION='0.5'
 
+
+class CommcareHqClientBuilder(object):
```
esoergel: I think you could replace this class with two functions (eg `build_client_with_auth` and `build_client_with_session`), since it doesn't look like the builder instance is needed.  That would simplify the calls anyways.  This is a style question, so feel free to disregard if you prefer this approach.

commcare_export/cli.py
esoergel: maybe something like `get_timestamp_params`?

commcare_export/cli.py
snopoke: Yes, session auth was originally the default but was replaced with Digest some time ago.

commcare_export/cli.py
snopoke: This got removed later but if it were empty then we'd get an error earlier on

snopoke:
Updated with feedback:
* remove builder in favor of simpler auth setup
* rename `get_since_until` to  `get_date_params`

PR #87, opened 2018-07-11
order by server_modified_on

snopoke:
With https://github.com/dimagi/commcare-hq/pull/21182


snopoke:
@jmtroth0 thanks for checking, the ordering does matter

snopoke:
I'll fix the build tomorrow

PR #88, opened 2018-07-11
misc

snopoke:


commcare_export/commcare_hq_client.py
```diff
@@ -165,7 +162,7 @@ def _key(self):
         return (self.username, self.apikey)
 
     def __eq__(self, other):
-        return self._key() == other.key()
+        return self._key() == other._key()
```
jmtroth0: This seems sort of against the typical "private method" convention

snopoke: :+1: updated: ca82826

PR #90, opened 2018-07-13
CLI + integration tests

snopoke:
This PR adds tests for the CLI code (previously untested) as well as an integration test that pulls data from HQ and saves it to a database.

Can review by commit :blowfish:


commcare_export/commcare_hq_client.py
```diff
@@ -52,7 +52,10 @@ def _get_auth(self, username, password, mode):
     @property
     def session(self):
         if self.__session == None:
-            self.__session = requests.Session(headers={'User-Agent': 'commcare-export/%s' % commcare_export.__version__})
+            self.__session = requests.Session()
+            self.__session.headers.update({
+                'User-Agent': 'commcare-export/%s' % commcare_export.__version__
+            })
         return self.__session
```
esoergel: huh, does `requests.Session()` specify some defaults that the old version overwrote?

.travis.yml
```diff
@@ -6,6 +6,8 @@ python:
 env:
   global:
     - MSSQL_SA_PASSWORD=Password@123
+    - secure: etv02uWtyy5P4DfyuHjFm5RDFc6WBHLsnIMC75VjDk8kxDgwV/lDbPYMh/hzfPHyskgA1EQbc8IfHlbZWFVV8jOTy+wvrVir/mw95AEyNyAL/TTSWvYfTvdCsxOSbY6vcGlJNfy6rc+y0h6QyuIknY0OhU8sTaRcQnvbFPnOz28=
+    - secure: aLj1bKtUF2CnAwG+yjiAjo39cKi9WHaonIwqsuhOx4McsD/xSz4QHv/6/XhXZ5KxKyxw1+PBl/mWo6gyrT5iHDRBPk5iJXqZAgQFS2ukZSv/tUBGL7bWzoO9YfoLuWllA33DCr3PiXAhkH53dTcor16UN9wXeCprBBSGjhpAxRQ=
```
esoergel: What's going on with these `secure` parameters?  Do they get digested into the username and api key referenced below in some way?

tests/conftest.py
```diff
@@ -47,12 +34,40 @@ def tear_down():
...
+        'admin_db': 'postgres'
+    },
+    {
+        'url': 'mysql+pymysql://travis@/%s?charset=utf8',
+    },
+    {
+        'url': 'mssql+pyodbc://SA:Password@123@localhost/%s?driver=ODBC+Driver+17+for+SQL+Server',
+        'admin_db': 'master'
+    }
+], ids=['postgres', 'mysql', 'mssql'])
```
esoergel: I'm new to `pytest`, would you mind explaining a bit about how this works?  From the docs it sounds like these `@pytest.fixture`s use dependecy injection to set up test data.  What is this `request` object - some built-in thing? Do the params specified here get set on that object?

snopoke: yes that's right. This will create the fixture 3 times, one for each set of params and any test that uses this fixture will get run once for each fixture.

The injection is based on the function name matching the test arg name.

esoergel: Cool, thanks!

Issue #94, opened 2018-08-01
Make it possible to change the query file without having to rebuild the entire database


snopoke:
Currently we use the md5 of the query file as the lookup for checkpoints but any modifications to the file change the md5 and cause a rebuild.

Particularly for large databases this may not be desired and users need more control over when a rebuild occurs.

In order to maintain the current workflow but still enable the new workflow a new command line argument should be added to provide a 'checkpoint key'. This key will then be used to maintain the checkpoint information and a change to the key will result in a full rebuild.

A change to the following arguments should also result in a rebuild (regardless of the checkpoint key):
* project space (--project arg)
* HQ instance being queried (--commcare-hq arg)

PR #98, opened 2018-08-14
use args in checkpoint

snopoke:
A change in some command line args should trigger a full rebuild. 

This adds `project` and `commcare-hq` args to the checkpoints.

Part of https://github.com/dimagi/commcare-export/issues/94

Can review per commit. First few commits are prefactors.


snopoke:
This tool uses an Excel file as it's config. When exporting data to SQL it keeps checkpoints of it's progress but if the md5 of the Excel file changes then it will restart the export from the beginning. This PR makes it restart if either of the `project` or `commcare url` change i.e. if they run the same export against a different project or different instance of HQ.

PR #101, opened 2018-08-24
Check for dictionary value before insert

bderenzi:
I'm not 100% sure this is the correct place for the fix, but it seems to be working so far.

### DIFF
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -428,6 +428,14 @@ class SqlTableWriter(SqlMixin, TableWriter):
             if val is None:
                 continue
 
+            # if it is a dictionary that has '#text', return that, otherwise set it to the empty string
+            if isinstance(val, dict):
+                if '#text' in val:
+                    row_dict[column] = val = val.get('#text')
+                else:
+                    row_dict[column] = val = ''                    
+                    continue
+            
             ty = self.best_type_for(val)
             if not column in columns:
                 logger.warn("Adding column '{}.{} {}'".format(table_name, column, ty))
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -428,14 +428,6 @@ class SqlTableWriter(SqlMixin, TableWriter):
             if val is None:
                 continue
 
-            # if it is a dictionary that has '#text', return that, otherwise set it to the empty string
-            if isinstance(val, dict):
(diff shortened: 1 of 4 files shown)


tests/test_writers.py
```diff
@@ -122,20 +122,24 @@ def test_insert(self, writer):
...
+                    ['bizzle', 1, 2, 3, 'hi'],
+                    ['bazzle', 4, 5, 6, 'hello'],
+                    ['bozzle', 7, 8, 9, {'#text': 'test_text','@case_type': 'person','@relationship': 'child','id': 'nothing'}],
+                    ['buzzle', 1, 8, 7, {'@case_type': '', '@relationship': 'child', 'id': 'some_id'}],
                 ]
             })
 
         # We can use raw SQL instead of SqlAlchemy expressions because we built the DB above
         with writer:
             result = dict([(row['id'], row) for row in writer.connection.execute('SELECT id, a, b, c FROM foo_insert')])
```
czue: I think you need to add `s` here to make the test pass

tests/test_writers.py
```diff
@@ -122,20 +122,24 @@ def test_insert(self, writer):
...
                 'name': 'foo_insert',
-                'headings': ['id', 'a', 'b', 'c'],
+                'headings': ['id', 'a', 'b', 'c', 's'],
                 'rows': [
-                    ['bizzle', 1, 2, 3],
-                    ['bazzle', 4, 5, 6],
+                    ['bizzle', 1, 2, 3, 'hi'],
+                    ['bazzle', 4, 5, 6, 'hello'],
+                    ['bozzle', 7, 8, 9, {'#text': 'test_text','@case_type': 'person','@relationship': 'child','id': 'nothing'}],
+                    ['buzzle', 1, 8, 7, {'@case_type': '', '@relationship': 'child', 'id': 'some_id'}],
```
snopoke: I actually think this should fail rather than return an empty result since it indicates that the source path is likely wrong

commcare_export/writers.py
```diff
@@ -428,6 +428,14 @@ def get_cols():
             if val is None:
                 continue
 
+            # if it is a dictionary that has '#text', return that, otherwise set it to the empty string
```
snopoke: I think a better place for this would be in `EmitterEnv`

PR #110, opened 2018-11-12
make 'id' column required for SQL export

snopoke:


commcare_export/excel_query.py
```diff
@@ -297,6 +297,12 @@ class SheetParts(namedtuple('SheetParts', 'name headings source body root_expr')
     def __new__(cls, name, headings, source, body, root_expr=None):
         return super(SheetParts, cls).__new__(cls, name, headings, source, body, root_expr)
 
+    @property
+    def columns(self):
+        return [
+            col.v for col in self.headings
```
calellowitz: Is `self.headings` the same as `sheet.headings` below

PR #111, opened 2018-11-25
empty commit for running tests

nickpell:


nickpell:
@snopoke It looks like the build is broken.  Could you advise?

PR #112, opened 2018-11-28
fix tests

snopoke:


tests/test_cli.py
```diff
@@ -47,7 +47,7 @@ def make_args(project='test', username='test', password='test', **kwargs):
 client = MockCommCareHqClient({
     'form': [
         (
-            {'limit': 1000, 'order_by': ['server_modified_on', 'received_on']},
```
nickpell: Why was this causing failures?

snopoke: I updated it but didn't update the mock

PR #118, opened 2019-03-01
don't set checkpoint based on the current time

czue:
@snopoke I'm not 100% sure about this but I think this is a potentially big oversight in the tool that could be causing lost/missing data.

What we observed when doing the local ingestions on ICDS was that the final checkpoint got set to today, even though we only ever saw data through September 2018 (or whenever).

By using the system's local time to set the checkpoint on the last request you are essentially saying that the server APIs can never get behind. This is obviously heavily exacerbated in the scenario we are using, but we know that ES gets behind all the time which I think could create gaps in the exports in steady state if people's exports are relying on local system time for the last checkpoint.

Note: this means the last checkpoint will not ever be set (assuming the tool terminates with empty API results) because [last_run will be empty here](https://github.com/dimagi/commcare-export/blob/master/commcare_export/checkpoint.py#L82-L83)

HT to @bderenzi for pointing this out.

### DIFF
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -151,6 +151,8 @@ class CommCareHqClient(object):
             since_date = paginator.get_since_date(batch)
             if since_date:
                 self._checkpoint_manager.set_batch_checkpoint(checkpoint_time=since_date)
+            else:
+                logger.warning('Failed to get a checkpoint date from a batch of data.')
 
 
 class MockCommCareHqClient(object):
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -149,7 +149,8 @@ class CommCareHqClient(object):
         from commcare_export.commcare_minilinq import DatePaginator
         if self._checkpoint_manager and isinstance(paginator, DatePaginator):
             since_date = paginator.get_since_date(batch)
-            self._checkpoint_manager.set_batch_checkpoint(checkpoint_time=since_date)
+            if since_date:
+                self._checkpoint_manager.set_batch_checkpoint(checkpoint_time=since_date)
 
 
 class MockCommCareHqClient(object):
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -141,7 +141,7 @@ class CommCareHqClient(object):
                     else:
                         more_to_fetch = False
 
-                self.checkpoint(paginator, batch)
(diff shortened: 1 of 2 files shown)


snopoke:
I think we should just make the `checkpoint_time` arg compulsory. It should always be called with the arg as of now:

For batch checkpoints it get's called the the last date from the batch:
```python
since_date = paginator.get_since_date(batch)
self._checkpoint_manager.set_batch_checkpoint(checkpoint_time=since_date)
```

For the final checkpoint it get's the date from the most recent checkpoint:
```python
last_run = self.get_time_of_last_checkpoint(log_warnings=False)
if last_run:
    self._set_checkpoint(dateutil.parser.parse(last_run), True)
```

czue:
ah it's probably not the final checkpoint but the last batch then.

`paginator.get_since_date(batch)` will return `None` if the batch is empty, but [the client still creates a checkpoint](https://github.com/dimagi/commcare-export/blob/master/commcare_export/commcare_hq_client.py#L144). 

You think it's better to make required and then remove the checkpoint call if there's no data in the batch?

snopoke:
Ah that makes sense.

> You think it's better to make required and then remove the checkpoint call if there's no data in the batch?

Yes 100%. If there is no data in the batch then we should not write a checkpoint.

Thanks for digging into that.

czue:
@snopoke updated. one side effect of this is that it will no longer create a record at all if there is no data in the API. But it does correctly create records if it has ever gotten data for that config file.

commcare_export/commcare_hq_client.py
```diff
@@ -141,15 +141,16 @@ def iterate_resource(resource=resource, params=params):
...
+                    self.checkpoint(paginator, batch)
                 
         return RepeatableIterator(iterate_resource)
 
     def checkpoint(self, paginator, batch):
         from commcare_export.commcare_minilinq import DatePaginator
         if self._checkpoint_manager and isinstance(paginator, DatePaginator):
             since_date = paginator.get_since_date(batch)
-            self._checkpoint_manager.set_batch_checkpoint(checkpoint_time=since_date)
+            if since_date:
```
snopoke: can you log a warning here if this is `None`

czue: added - feel free to suggest text changes

czue:
@snopoke do you think we should cut a release for this?

snopoke:
why don't you test it out with ICDS for a bit first

czue:
k, will do.

PR #120, opened 2019-05-29
fix output warnings

snopoke:


tests/conftest.py
```diff
@@ -15,6 +15,13 @@
 logging.getLogger().addHandler(logging.StreamHandler())
 
 
+def pytest_configure(config):
+    config.addinivalue_line("markers", "dbtest: mark test that requires database access")
+    config.addinivalue_line("markers", "postgres: mark PostgreSQL test")
+    config.addinivalue_line("markers", "mysql: mark MySQL test")
+    config.addinivalue_line("markers", "mssql: mark MSSQL test")
```
mkangia: is this needed twice?

snopoke: one is `mysql` the other is `mssql`

Issue #119, opened 2019-05-22
Query files with multiple sheets share the same checkpoint


snopoke:
When a query file has multiple sheets that use different data sources or have different filters the tool processes each sheet in sequence. After fetching each batch the checkpoint is updated with the dates from that batch so successive sheets overwrite the checkpoint dates of prior ones.
```
Fetch for sheet 1
Update checkpoint to A

Fetch for sheet 2
Update checkpoint to B
```

This means that between successive runs it could be possible to miss data.

To resolve this each checkpoint should be unique for the table being written to. The MD5 should also be per sheet and not for the entire query file. MD5 can be calculated based on the generated minilinq query JSON.

PR #121, opened 2019-05-29
checkpoint per table

snopoke:
fixes https://github.com/dimagi/commcare-export/issues/119

This changes the way checkpoints are recorded. Previously a single checkpoint would be used per query file however the query file could have multiple sheets in it, each representing a different export table and different datasource.

During the export the tool processes each sheet in turn and updates the checkpoint as it goes. The result of this is that the checkpoint for earlier tables get's overridden by the updates from later tables.

To address this I've added a `table_name` column to the checkpoint. To avoid a lot of hackery in passing around table names I instead bind a checkpoint manager into the context for each table:

```python
Bind(
    'checkpoint_manager', Apply(Reference('get_checkpoint_manager'), Literal(["table1"])),
    Emit(
            source = Map(
                source=Apply(Reference("api_data"), Literal("form"), Reference('checkpoint_manager')),
                body = ...
            )
     )
)
```
This checkpoint manager get's passed to the `api_data` function and used to record the checkpoints. This is somewhat analogous to
```python
with get_checkpoint_manager(['table1']) as cm:
    data_iterator = api_data('form', cm)
```


commcare_export/checkpoint.py
```diff
@@ -167,18 +222,84 @@ def list_checkpoints(self, limit=20):
...
+            raise Exception("Not tables set in checkpoint manager")
+
+
+class CheckpointManagerWithSince(object):
+    def __init__(self, manager, since):
+        self.manager = manager
+        self.since_param = since
+
+    def set_checkpoint(self, checkpoint_time, is_final=False):
+        if self.manager:
```
millerdev: Are there known scenarios where `set_checkpoint()` would be called when `self.manager is None`? If not, what do you think of raising an exception here rather than returning silently?

Prior to this PR, it looks like a call to `set_checkpoint()` always set a checkpoint, but this makes it possible that it would be a no-op. That seems like it could be bad (if there was a bug that caused `self.manager` to be false-y).

millerdev: I'm not blocking on this, but it sounds like the checkpoint manager is maybe doing more than one thing[<sup>1</sup>](#do-one-thing) if it sometimes is used to set checkpoints and other times is only the source of the `since` value.


<sup><a name="do-one-thing">1</a></sup> ref. "Do One Thing" concept from [Clean Code](https://www.amazon.com/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882), which we're reading in book club right now.

snopoke: I have already applied this concept:
* `CheckpointManager` is solely responsible for managing checkpoints.
* `CheckpointManagerWithSince` is a container for the `since` value and checkpoint manager specific to the current table export.
* `CheckpointProvider` is responsible for creating new the creating a new `CheckpointManagerWithSince` when requested with the `since` value and the checkpoint manager.

The `since` value and the checkpoints are inextricably tied together because unless specified at the command line the `since` value comes from the checkpoints which is why it makes sense to be part of the checkpoint manager: https://github.com/dimagi/commcare-export/pull/121/files#diff-a29d76e0f5317a0f68900178498d8141R314

Having said all that I'm open to suggestions for improvement. I did got through several iterations with this and originally didn't have the `CheckpointManagerWithSince`. I introduced that class so that the checkpoint manager with tables set would only live inside that class which in turn would only live inside the context of the of `Bind`.

commcare_export/checkpoint.py
```diff
@@ -124,25 +135,38 @@ def get_time_of_last_checkpoint(self, log_warnings=True):
...
+                        table_run = self._get_last_checkpoint(
+                            session, table_name=table,
+                            query_file_md5=self.query_md5, key=self.key
+                        )
+                table_runs.append(table_run)
+
+        sorted_runs = list(sorted(filter(None, table_runs), key=attrgetter('time_of_run')))
+        return sorted_runs[0] if sorted_runs else None
+
+    def _get_last_checkpoint(self, session, *criterion_filters, **kwarg_filters):
```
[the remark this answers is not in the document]
snopoke: At one point I was using these to do an `in` filter: `Checkpoint.table_name.in_(self.table_names)` but moved away from that approach. I'll remove that for now since it's not used.

commcare_export/checkpoint.py
millerdev: You highlighted what I'm getting at by using the word "and" in the description of `CheckpointManagerWithSince`:

> a container for the `since` value _and_ checkpoint manager specific to the current table export.

(emphasis added)

Note: previously I had referenced "Do One Thing," a concept Uncle Bob uses when talking about functions, when I actually meant to reference "Single Responsibility Principle," which he uses when talking about classes. Here's a relevant quote from the chapter on Classes near where he's talking about the Single Responsibility Principle:

> The name of a class should describe what responsibilities it fulfills. In fact, naming is probably the first way of helping determine class size. If we cannot derive a concise name for a class, then it’s likely too large. The more ambiguous the class name, the more likely it has too many responsibilities. For example, class names including weasel words like `Processor` or `Manager` or `Super` often hint at unfortunate aggregation of responsibilities. We should also be able to write a brief description of the class in about 25 words, without using the words “if,” “and,” “or,” or “but.”

I'm going to let this one go. The PR fixes the issue that prompted the change, and I don't have enough context here to easily suggest how to improve it.

snopoke: I'm aware of the principle but I disagree that it applies here. It would be like applying it to a DB access class that reads "and" writes.

As I already said, the `since` value is inextricably tied to the checkpoints. If I rename it like this does it make it clearer for you:

```python
class CheckpointInterface(object):
    def get_initial_checkpoint_value(self)
        pass

    def set_checkpoint(self, value)
        pass
```

millerdev: Does `get_initial_checkpoint_value()` return what used to be `since_param`? Will that interface be implemented by both `CheckpointManagerWithSince` and `CheckpointManager`? I guess it would be helpful to see the implementation(s) of that interface.

snopoke: Yes, it would return the same value. I'll try and find some time this week to refactor these classes.

PR #122, opened 2019-07-03
use text / nvarchar type for ID column for PG and MSSQL

snopoke:
https://dimagi-dev.atlassian.net/browse/HI-707


snopoke:
ProgrammingError: (pyodbc.ProgrammingError) ('42000', u"[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Column 'id' in table 'foo_insert' is of a type that is invalid for use as a key column in an index. (1919) (SQLExecDirectW)")

Issue #107, opened 2018-10-08
Switch to using date added to ES


bderenzi:
There is a bug where any unprocessed form that is in the backlog for ES will never be sync'd to a just-completed sync because we use form submission date as opposed to the data added to ES.

The proposal is to switch to using date-added-to-ES so that backlog forms will be sync'd as well.

snopoke:
Another reason to do this is that it will prevent the tool from getting stuck when there is more than 'batch-size' docs with the same modified on date e.g. cases updated by a large case import.

snopoke:
https://github.com/dimagi/commcare-hq/pull/24393

snopoke:
Fixed by https://github.com/dimagi/commcare-export/pull/159

PR #124, opened 2019-07-05
paginate forms and cases by indexed_on

snopoke:
Fixes https://github.com/dimagi/commcare-export/issues/107

Depends on https://github.com/dimagi/commcare-hq/pull/24741

### DIFF
--- a/commcare_export/commcare_minilinq.py
+++ b/commcare_export/commcare_minilinq.py
@@ -31,52 +31,9 @@ class SimpleSinceParams(object):
         return params
 
 
-class FormFilterSinceParams(object):
-    def __call__(self, since, until):
-        range_expression = {}
-        if since:
-            range_expression['gte'] = since.isoformat()
-
-        if until:
-            range_expression['lte'] = until.isoformat()
-
-        server_modified_missing = {"missing": {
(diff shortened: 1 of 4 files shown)


snopoke:
I was thinking about this some more and if we use `indexed_on` it will mean that an ES reindex (using the current reindex tools we have) would cause the export tool to re-download all the data. I'm not sure how big of an issue that is but I think I'll leave this for now until I've had a chance to discuss it more.

snopoke:
Build broken because integration test with HQ is now filtering on `indexed_on` so would need to change the date ranges etc.

snopoke:
One option to solve the above could be to sort and filter by `indexed_on` as well as `received_on`

esoergel:
> One option to solve the above could be to sort and filter by indexed_on as well as received_on

I don't follow - do you mean both together in some way?  Or alternately?

snopoke:
@esoergel we could sort by both `received_on` and `indexed_on` and then also add both of those to the next page filter. This would solve the issue of having a large set of forms with the same `received_on` date cause the tool to get stuck in a loop which is the 2nd issue described in https://github.com/dimagi/commcare-export/issues/107

snopoke:
https://github.com/dimagi/commcare-export/pull/159

PR #127, opened 2019-10-24
modify batch size when cycle detected

snopoke:
This is a workaround to the issue described here https://github.com/dimagi/commcare-export/issues/107#issuecomment-435052926

The other PR dealing with this is still a better long term solution but requires more effort: https://github.com/dimagi/commcare-export/pull/124

### DIFF
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -8,6 +8,8 @@ import requests
 from requests.auth import AuthBase
 from requests.auth import HTTPDigestAuth
 
+from commcare_export.exceptions import DataExportException
+
 AUTH_MODE_PASSWORD = 'password'
 AUTH_MODE_APIKEY = 'apikey'
 
@@ -23,8 +25,8 @@ from commcare_export.repeatable_iterator import RepeatableIterator
 
 logger = logging.getLogger(__name__)
 
-LATEST_KNOWN_VERSION='0.5'
-
+LATEST_KNOWN_VERSION = '0.5'
+MAX_BATCH_SIZE = 1000
 
 def on_backoff(details):
     _log_backoff(details, 'Waiting for retry.')
@@ -52,7 +54,7 @@ class CommCareHqClient(object):
     """
 
     def __init__(self, url, project, username, password,
-                 auth_mode=AUTH_MODE_PASSWORD, version=LATEST_KNOWN_VERSION, checkpoint_manager=None):
+                 auth_mode=AUTH_MODE_PASSWORD, version=LATEST_KNOWN_VERSION):
         self.version = version
         self.url = url
         self.project = project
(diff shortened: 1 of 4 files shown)


commcare_export/commcare_hq_client.py
```diff
@@ -117,31 +119,48 @@ def iterate_resource(resource=resource, params=params):
...
             while more_to_fetch:
+                current_limit = params.get('limit')
                 batch = self.get(resource, params)
                 if not total_count or total_count == 'unknown' or fetched >= total_count:
                     total_count = int(batch['meta']['total_count']) if batch['meta']['total_count'] else 'unknown'
                     fetched = 0
 
-                fetched += len(batch['objects'])
+                new_in_batch = [obj for obj in batch['objects'] if obj['id'] not in last_batch_ids]
+                last_batch_ids = {obj['id'] for obj in batch['objects']}
```
esoergel: Checking my understanding here:
Besides the initial request, each subsequent request is going to be a strict superset of results, therefore there's no need to keep track of `last_batch_ids` beyond the most recent request.

Or is there a possibility of say, non-deterministic ordering which could result in objects in batch 1 that aren't also in batch 2?

For instance, this should work fine:
Batch 1 - fetch 5 items starting at date A: [A1, A2, A3, B1, B2]
Batch 2 - fetch 5 items starting at date B: [B1, B2, B3, B4, B5]
Batch 3 - fetch 7 items starting at date B: [B1, B2, B3, B4, B5, B6, B7]

But if the ordering is inconsistent beyond date, there could be issues:
Batch 1 - fetch 5 items starting at date A: [A1, A2, A3, B1, B2]
Batch 2 - fetch 5 items starting at date B: [B2, B3, B4, B5, B6]  # missing B1!
Batch 3 - fetch 7 items starting at date B: [B1, B2, B3, B4, B5, B6, B7]  # B1 will be yielded a second time

snopoke: Interesting question. I was not able to determine from any documentation whether or not the sort order would deterministic however in practice we've seen it to be so.

esoergel: That's a good idea.  Seems a bit clearer of an approach too.  Though this would still need to exclude duplicates when fetching overlapping data.

kaapstorm: > Another possible way to break the cycle would be to compare the filter parameters. If they are identical then we would increase the batch size.

I think this is a great idea. Not only would it work regardless of sorting, it would also pre-empt the problem. In the Ethan's example, in Batch 2, when we get to B5 we already know we must increase the batch size. If we keep going through B6 and B7 to C1, then Batch 3 will start at C1 at worst, and D1 at best.

We would want a max batch size, and if we reach it we would know we have an unrecoverable problem, so probably best to make it high.

PR #129, opened 2019-12-20
drop option to read json query from stdin

snopoke:
I don't think this is ever used and if the user forgets to pass the `query` arg then the tool hangs forever waiting for stdin. You can still use a json query by having it in a file: `--query=query.json`

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -52,7 +52,7 @@ class Argument(object):
 CLI_ARGS = [
         Argument('version', default=False, action='store_true',
                  help='Print the current version of the commcare-export tool.'),
-        Argument('query', help='JSON or Excel query file. If omitted, JSON string is read from stdin.'),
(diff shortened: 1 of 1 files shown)


commcare_export/cli.py
```diff
@@ -52,7 +52,7 @@ def add_to_parser(self, parser, **additional_kwargs):
 CLI_ARGS = [
         Argument('version', default=False, action='store_true',
                  help='Print the current version of the commcare-export tool.'),
-        Argument('query', help='JSON or Excel query file. If omitted, JSON string is read from stdin.'),
+        Argument('query', required=True, help='JSON or Excel query file. If omitted, JSON string is read from stdin.'),
```
kaapstorm: I think you need to drop the sentence "If omitted, JSON string is read from stdin."

snopoke:
```suggestion
        Argument('query', required=True, help='JSON or Excel query file'),
```

PR #131, opened 2020-03-10
support locations in DET

czue:
This works locally. Any reason not to add it?

Also, do we know why users / apps / web users aren't documented at all?


snopoke:
> Also, do we know why users / apps / web users aren't documented at all?

Just never been done.

PR #133, opened 2020-04-15
Users and locations queries

solleks:
This pull request has multiple changes.

1) A small change to allow MSSQL tests to run on Mac OS for me.
2) Add instructions to README for installing MSSQL database on Mac OS.
3) Implement the --users and --locations options to automatically export user data and location data, including some location type fields. See this design doc (https://docs.google.com/document/d/1XVVxAdlSbEMMMUEhzzJGVx83LBjJblwTNdvc1Pbtsl4/edit?usp=sharing) for details.

Automatically connecting exported tables to the user and location data will be a separate change, I expect.

Please let me know if you'd like this pull request split up into multiple pieces or if I need to add some more tests.


README.md
```diff
@@ -73,10 +73,13 @@ The basic usage of the command-line tool is with a saved Excel or JSON query (se
 $ commcare-export --commcare-hq <URL or alias like "local" or "prod"> \
                   --username <username> \
                   --project <project> \
-                  --version <api version, defaults to latest known> \
+                  --api-version <api version, defaults to latest known> \
+                  --version <print current version> \
```
czue, line 77: :+1: thanks for this fix

commcare_export/cli.py
```diff
@@ -207,19 +218,98 @@ def _get_checkpoint_manager(args):
...
+        if not query:
+            print('Query file not found: %s' % args.query)
+            return EXIT_STATUS_ERROR
+        else:
+            query_tables = query.emitted_tables()
+
+    if args.users:
+        if builtin_users_query.TABLE_NAME in query_tables:
+            print('--users flag should not be used with a query that writes '
+                  'to the table "commcare_users"')
```
czue: nice

czue:
fyi tests are failing. not sure if related or not.
```
=================================== FAILURES ===================================
_____ TestCLIIntegrationTests.test_write_to_sql_with_checkpoints[postgres] _____
self = <test_cli.TestCLIIntegrationTests object at 0x7f25aa0d8410>
writer = <commcare_export.writers.SqlTableWriter object at 0x7f259cc886d0>
checkpoint_manager = <commcare_export.checkpoint.CheckpointManager object at 0x7f25a35a1450>
caplog = <_pytest.logging.LogCaptureFixture object at 0x7f25a35a1e50>
    def test_write_to_sql_with_checkpoints(self, writer, checkpoint_manager, caplog):
        with open('tests/009_expected_form_data.csv', 'r') as f:
            reader = csv.reader(f)
            expected_form_data = list(reader)[1:]
    
>       _pull_data(writer, checkpoint_manager, 'tests/009_integration.xlsx', '2012-01-01', '2012-08-01')
tests/test_cli.py:325: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/test_cli.py:299: in _pull_data
    username=os.environ['HQ_USERNAME'],
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = {'TRAVIS_UID': '2000', 'rvm_version': '1.29.7 (latest)', 'LC_CTYPE': 'en_US.UTF-8', 'TRAVIS': 'true', 'PERLBREW_SHELLR...'HISTCONTROL': 'ignoredups:ignorespace', 'PWD': '/home/travis/build/dimagi/commcare-export', 'TRAVIS_OS_NAME': 'linux'}
key = 'HQ_USERNAME'
    def __getitem__(self, key):
        if key in self.data:
            return self.data[key]
        if hasattr(self.__class__, "__missing__"):
            return self.__class__.__missing__(self, key)
>       raise KeyError(key)
E       KeyError: 'HQ_USERNAME'
../../../virtualenv/python2.7.15/lib/python2.7/UserDict.py:40: KeyError
```

commcare_export/builtin_location_types_query.py
```diff
@@ -0,0 +1,155 @@
...
+    headings = location_types[TABLE_NAME]['headings']
+    rows = location_types[TABLE_NAME]['rows']
+    for row in rows:
+        resource_uri = None
+        row_dict = {}
+        assert len(row) == len(headings)
+        for i in range(0, len(headings)):
+            row_dict[headings[i]] = row[i]
+            if headings[i] == 'resource_uri':
+                resource_uri = row[i]
```
snopoke, line 137:
```suggestion
        row_dict = dict(zip(headings, row))
        resource_uri = row_dict.get('resource_uri')
```

tests/test_cli.py
```diff
@@ -62,49 +62,219 @@ def make_args(project='test', username='test', password='test', **kwargs):
...
+        args = make_args(
+            output_format='json',
+            locations=True
+        )
+        writer = JValueTableWriter()
+        with mock.patch('commcare_export.cli._get_writer', return_value=writer):
+            main_with_args(args)
+
+        for table in EXPECTED_LOCATIONS_RESULTS:
+            assert writer.tables[table['name']] == table
```
snopoke, line 264: These lines are common to all tests and could be extracted to a util function:
```suggestion
        self._test_cli(args, EXPECTED_LOCATIONS_RESULTS)
```

commcare_export/cli.py
```diff
@@ -207,19 +218,99 @@ def _get_checkpoint_manager(args):
...
+            return EXIT_STATUS_ERROR
+
+        if not query:
+            print('Query file not found: %s' % args.query)
+            return EXIT_STATUS_ERROR
+        else:
+            query_tables = query.emitted_tables()
+
+    if args.users:
+        if builtin_users_query.TABLE_NAME in query_tables:
```
snopoke, line 292: rather than doing this and having to add all the `emitted_tables` function definitions I think we should just prevent exporting to this table via an excel query completely. This should be easy to do in this function by checking `sheet.name` against a blacklist: https://github.com/dimagi/commcare-export/blob/77d4a9e0ad61878bd9aaf35efa00a7922c63b415/commcare_export/excel_query.py#L451-L459

commcare_export/cli.py
```diff
@@ -207,19 +218,99 @@ def _get_checkpoint_manager(args):
...
+                print("\nAuthentication failed. Please check your credentials.")
+                return None
+            else:
+                raise
+        except KeyboardInterrupt:
+            print('\nExport aborted')
+            return None
+
+
+def location_type_map(args):
```
snopoke, line 240: this should be done lazily as part of the main query evaluation. I also don't think it needs to use a minilinq query, you could just use the api client directly to get the API json output and do the processing in python which would be much less verbose. You could do that by passing the `api_client` into the LocationInfoProvider class:

```python
class LocationInfoProvider:
    def __init__(api_client):
        self._api_client = api_client
        self._location_types = None

    @property
    def location_types(self):
        if self._location_types is None:
            self._location_types = get_location_types(self.api_client)
        return self._location_types

def get_location_types(api_client):
    response = api_client.get('location_type')
    .....
```

commcare_export/cli.py
```diff
@@ -207,19 +218,99 @@ def _get_checkpoint_manager(args):
...
-        print(e.message)
+    if args.query is None and args.users is False and args.locations is False:
+        print('At least one the following arguments is required: '
+              '--query, --users, --locations')
         return EXIT_STATUS_ERROR
 
-    if not query:
-        print('Query file not found: %s' % args.query)
-        return EXIT_STATUS_ERROR
+    query = None
```
snopoke, line 276: this code below could be put in a function:
```
try:
    query = get_queries(args)
except DataExportException as e:
    print(e.message)
    return EXIT_STATUS_ERROR
```

commcare_export/cli.py
```diff
@@ -207,19 +218,99 @@ def _get_checkpoint_manager(args):
...
+            query_tables = query.emitted_tables()
+
+    if args.users:
+        if builtin_users_query.TABLE_NAME in query_tables:
+            print('--users flag should not be used with a query that writes '
+                  'to the table', builtin_users_query.TABLE_NAME)
+            return EXIT_STATUS_ERROR
+        # Add user data to query
+        user_query = MiniLinq.from_jvalue(builtin_users_query.jvalue)
+        query = MiniLinq.combine(query, user_query)
```
snopoke, line 298: I don't think this `combine` function is necessary. It's simpler to do something like this: https://github.com/dimagi/commcare-export/blob/77d4a9e0ad61878bd9aaf35efa00a7922c63b415/commcare_export/excel_query.py#L458-L459

You could also then get rid of `List.concatenate`

commcare_export/builtin_locations_query.py
```diff
@@ -0,0 +1,231 @@
+
+TABLE_NAME = "commcare_locations"
+
+# A JSON format MiniLinq query for internal CommCare location table.
+# It reads every field produced by the /location/ API endpoint and
+# writes the data to a table named "commcare_locations" in a database.
+jvalue = {
```
snopoke, line 7: one issue with having this as the full query is that if there are changes to how we compile queries in the future this may not continue to work without applying the appropriate changes. One way to 'future proof' this would be to store this as a set of `SheetParts` and then use `compile_queries` to convert that into the final query. I think that would also make this much less verbose:

```python

class Column:
    def __init__(name, source, map_function=None, *extra_args):
        ...

    @property
    def output_field(self):
        if not self.map_function:
            return Reference(self.source)
        else:
            return Apply(Reference(self.map_function), Reference(self.source), *self.extra_args)

columns = [
    Column('id', 'id'),
    Column('created_at', 'created_at', 'str2date'),
    ...
    Column('location_type_administrative', 'location_type', 'get_location_info, 'administrative')
    ...
]
source = Apply(Reference("api_data"), Literal(data_source), Reference('checkpoint_manager'))
part = SheetParts(TABLE_NAME, [c.name for c in columns], source, List([c.output_field for c in columns], None)
query = compile_queries([part], None, False)

**Comment by solleks** [comment]
Thank you for all of the comments. I plan to make all of the suggested changes, but haven't finished implementing them yet.

The problem with the integration tests seems to be due to the use of [secure environment variables in Travis CI](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml). They are intentionally undefined in a pull request like this so no one can steal them just by making a pull request.

**Comment by snopoke** [comment]
I don't think there's a way around the integration test issue but once the PR is ready to merge I can run them locally - or you could too if you have access to the domain on CommCare HQ

File: commcare_export/builtin_location_types_query.py (hunk 3 above)
Comment (solleks) [reply commcare_export/builtin_location_types_query.py]: Done, and the code moved to location_info_provider.py.

File: tests/test_cli.py (hunk 4 above)
Comment (solleks) [reply tests/test_cli.py]: Done.

File: commcare_export/cli.py (hunk 5 above)
Comment (solleks) [reply commcare_export/cli.py]: Thanks for the suggestion. I made adding tables to the blacklist be a call from builtin_queries.py to avoid circular imports into excel_query.py.

File: commcare_export/cli.py (hunk 6 above)
Comment (solleks) [reply commcare_export/cli.py]: Much better. Done.

File: commcare_export/cli.py (hunk 7 above)
Comment (solleks) [reply commcare_export/cli.py]: Done.

File: commcare_export/builtin_locations_query.py (hunk 9 above)
Comment (solleks) [reply commcare_export/builtin_locations_query.py]: I like it. Done.

File: commcare_export/cli.py (hunk 8 above)
Comment (solleks) [reply commcare_export/cli.py]: I tried to make this change but ran into a problem and backed it out.

The problem appears in tests like "test_cli_table_plus_users" where the Excel query emits to two tables, which gave a MiniLinq query like List(List(table1, table2), users). The test failed by not finding the expected output for one of the Excel query produced tables.

So it seems that a nested List query does not behave like a flattened List query. Are they expected to behave the same?

Please let me know if I missed something obvious here.

Comment (snopoke) [reply commcare_export/cli.py]: I tried doing this locally and it worked fine and the tests pass:

```python
def get_queries(args, writer):
    query_list = []
    if args.query is not None:
        query = _get_query(args, writer)
        if not query:
            raise MissingQueryFileException(args.query)
        query_list.append(query)

    if args.users:
        # Add user data to query
        query_list.append(builtin_queries.users_query)

    if args.locations:
        # Add location data to query
        query_list.append(builtin_queries.locations_query)

    return List(query_list) if len(query_list) > 1 else query_list[0]
```

solleks: I made this change, even though it causes two of the new tests to fail locally (test_cli_table_plus_users and test_cli_table_plus_locations). I'm expecting them to fail in Travis CI as well, although all I can see right now is a failed test with an empty log.

Since the tests always failed for me locally, I tracked down the cause. The value of 'lazy_result' in evaluate_query in cli.py was a nested list such as [[RepeatableIterator, None], None] and that function only traversed the top level list, never iterating over the nested RepeatableIterator. Changing that code to perform a recursive traversal fixed the problem for me, but I didn't want to make that change until we agree that there is actually a problem.

snopoke: fine to make that recursively traverse the result.

solleks: OK, I added the recursive traversal and now only see the integration tests failing in Travis CI, as expected.

PR #136, opened 2020-05-18
oracle column

calellowitz:
This changes the default oracle text column to an NVARCHAR2(4000)

### DIFF
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -358,7 +358,7 @@ class SqlTableWriter(SqlMixin, TableWriter):
             elif self.is_mssql:
                 return sqlalchemy.NVARCHAR(collation=self.collation)
             if self.is_oracle:
-                return sqlalchemy.UnicodeText(collation=self.collation)
+                return sqlalchemy.Unicode(collation=self.collation)
             else:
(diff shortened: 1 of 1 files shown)


commcare_export/writers.py
```diff
@@ -358,7 +358,7 @@ def best_type_for(self, val):
             elif self.is_mssql:
                 return sqlalchemy.NVARCHAR(collation=self.collation)
             if self.is_oracle:
-                return sqlalchemy.UnicodeText(collation=self.collation)
+                return sqlalchemy.Unicode(4000, collation=self.collation)
```
ctsims, line 361: shouldn't this be NVARCHAR(4000? Just looking at the sqlalchemy one above...)

calellowitz: According to sqlalchemy, `Unicode` is converted under the hood on oracle to `VARCHAR2`, just as `UnicodeText` was converted to `CLOB` https://docs.sqlalchemy.org/en/13/dialects/oracle.html#unicode-specific-column-datatypes. I don't have a good sense of what is the preferred way to specify that, but this seems a bit more flexible and inline with our other usages.

PR #137, opened 2020-05-19
use stderr instead of stdout for error messages

czue:
This makes it easier to programatically access these (which is used by commcare sync)

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -120,7 +120,7 @@ def main(argv):
         exit(0)
 
     if not args.project:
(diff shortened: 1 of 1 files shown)


czue:
> I think the 'pythonic' way to print to stderr is print("message", file=sys.stderr) 

I had just copied it [from elsewhere in the file](https://github.com/dimagi/commcare-export/blob/master/commcare_export/cli.py#L103-L104) but that does seem better. let me test/update

czue:
updated @millerdev @ctsims

PR #140, opened 2020-05-26
Simplified --with-organization change

solleks:
Here is a simplified version of the change to add organizational data to the export tool. It avoids creating views in a database to reduce brittleness in the face of changes.

The features of the simplified version are:
- add commcare_userid to Excel query specs when --with-organization is used.
- add one column to the commcare_locations table for each organization level and populate it with the location_id of a location's ancestor at that level.
- document the typical pattern for joining user, location and form data together in SQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,8 @@ $ commcare-export --commcare-hq <URL or alias like "local" or "prod"> \
                   --output-format <csv, xls, xlsx, json, markdown, sql> \
                   --output <file name or SQL database URL> \
                   --users <export data about project's mobile workers> \
-                  --locations <export data about project's location hierarchy>
+                  --locations <export data about project's location hierarchy> \
+                  --with-organization <export users, locations and joinable form or case tables>
 ```
 
 See `commcare-export --help` for the full list of options.
@@ -127,9 +128,11 @@ User and Location Data
 ----------------------
 
 The --users and --locations options export data from a CommCare project that
-can be joined with form and case data. Specifiying the --users option will
-export an additional table named 'commcare_users' containing the following
-columns:
+can be joined with form and case data. The --with-organization option does all
+of that and adds a field to Excel query specifications to be joined on.
+
+Specifiying the --users option or --with-organization option will export an
+additional table named 'commcare_users' containing the following columns:
 
 Column                           | Type | Note
 ------                           | ---- | ----
@@ -150,8 +153,8 @@ username                         | Text |
 The data in the 'commcare_users' table comes from the [List Mobile Workers
 API endpoint](https://confluence.dimagi.com/display/commcarepublic/List+Mobile+Workers).
 
-Specifying the --locations option will export an additional table named
-'commcare_locations' containing the following columns:
+Specifying the --locations option or --with-organization options will export
+an additional table named 'commcare_locations' containing the following columns:
 
 Column                       | Type | Note
 ------                       | ---- | ----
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -172,15 +175,58 @@ location_type_administrative | Text |
...
 given a query specification that writes to either of them.
 
+The export tool will write all users to 'commcare_users' and all locations to
+'commcare_locations', overwriting existing rows with current data and adding
+rows for new users and locations. If you want to remove obsolete users or
+locations from your tables, drop them and the next export will leave only
+the current ones. If you modify your organization to add or delete levels,
+you will change the columns of the 'commcare_locations' table and it is
+very likely you will want to drop the table before exporting with the new
+organization.
```
millerdev, line 228: 👍 good docs

commcare_export/excel_query.py
```diff
@@ -258,14 +258,47 @@ def compile_source(worksheet):
...
+    extend_fields = True
+
+    required_column = column_enforcer.column_to_require(data_source)
+    if required_column is None:
+        extend_fields = False
+    else:
+        for i in range(len(output_headings)):
+            if output_headings[i].value == required_column.name.v:
+                if isinstance(output_fields[i], Reference) and \
+                   output_fields[i].ref == required_column.source:
```
millerdev, line 280: nit: recommended indentation it two-levels for wrapped continuation
```suggestion
                if isinstance(output_fields[i], Reference) and \
                        output_fields[i].ref == required_column.source:
```

README.md
```diff
@@ -150,8 +153,8 @@ username                         | Text |
 The data in the 'commcare_users' table comes from the [List Mobile Workers
 API endpoint](https://confluence.dimagi.com/display/commcarepublic/List+Mobile+Workers).
 
-Specifying the --locations option will export an additional table named
-'commcare_locations' containing the following columns:
+Specifying the --locations option or --with-organization options will export
+an additional table named 'commcare_locations' containing the following columns:
 
 Column                       | Type | Note
 ------                       | ---- | ----
```
millerdev, line 160: Looks like `parent` column is missing here. Was that intentionally left out? Since the column has been around for a while, maybe it makes sense to simply document that the value of `parent` is a resource URI rather than changing the column name as in https://github.com/dimagi/commcare-export/pull/138/commits/6a7374c1f8cb1185b13c6833b15f1ca6d95487c5?

README.md
solleks: Thanks

README.md
solleks: OK, I'll add parent with a coment.

solleks:
Thanks for the quick review.

I saw some Travis CI failures with Python 2 that I fixed.

There is a known issue that two integration tests fail when run for pull requests from outside of Dimagi, due to Travis hiding encrypted environment variables, but those should be the only failures. Those tests pass when run from my machine.

README.md
czue: +1

czue:
@millerdev is this good to merge from your perspective? a few small changes since your last approval.

czue:
I think tests will continue to fail due to this problem:

> There is a known issue that two integration tests fail when run for pull requests from outside of Dimagi, due to Travis hiding encrypted environment variables, but those should be the only failures. Those tests pass when run from my machine.

On @solleks 's [last PR](https://github.com/dimagi/commcare-export/pull/133) we just merged it after they had passed locally (which I think he's done, yeah?)

millerdev:
I'm not sure I'm understanding. Are you saying Travis will always fail as long as those problematic tests exist? That doesn't seem like a good state to be in. We should either find a way to rewrite those tests so they pass or skip them so Travis is not failing.

czue:
> Are you saying Travis will always fail as long as those problematic tests exist? 

Only if the PR author is not a member of the Dimagi org. See Simon's comment here: https://github.com/dimagi/commcare-export/pull/133#issuecomment-614484064

millerdev:
Now I get it.

> Encrypted variables are not available to untrusted builds such as pull requests coming from another repository.

That makes sense. Sorry for not reading all the context. Good to merge if the tests have passed locally.

solleks:
All tests pass locally, so I will try out my new permissions to merge this PR.

solleks:
I updated the Confluence Wiki page to cover the --with-organization option.

czue:
woohoo! Thanks @solleks :raised_hands:

czue:
(I think we should ship a new release with this and data type support once that's merged... I can handle that once i get it through)

PR #145, opened 2020-06-08
Avoid fatal errors for unknown locations or types

solleks:
Try to fix a DET bug reported via Slack:
https://app.slack.com/client/T02QG622C/CGL7960TV

The line number was missing, but most likely, a previously unseen location resource URI was used as a key in the location_data table. It was probably the parent location of some location that was in the table. That contradicts my assumption that all locations and their parents would be provided by the api_client iterating over the 'location' endpoint.

This fix just changes fatal KeyErrors into warnings, so customers can keep using the --locations feature.

### DIFF
--- a/commcare_export/location_info_provider.py
+++ b/commcare_export/location_info_provider.py
@@ -1,5 +1,9 @@
+import logging
+
 from commcare_export.misc import unwrap_val
 from commcare_export.commcare_minilinq import SimplePaginator
+
+logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


solleks:
The integration tests succeed locally for me, although they fail in Travis CI, as expected.

czue:
@solleks are you able to push to a branch in the dimagi repo so that the integration tests pass? Or you don't have that permission? (I can try to fix on github if the latter)

czue:
will just merge for now since this looks unlikely to cause integration test issues

PR #149, opened 2020-06-12
Raise exception on repeated requests with no progress

solleks:
Here is one short-term solution to the problem of batches with all the same timestamp - stop and suggest that the user increase the batch size.

Related doc: https://docs.google.com/document/d/1YnS9mN2gxG-OBREoxLNgpWwIJJgIFW-4NU2QznKG0XM/edit?ts=5ede8253#

I asked some questions in email about the alternative solution of checking for repeated results.

### DIFF
--- a/tests/test_commcare_hq_client.py
+++ b/tests/test_commcare_hq_client.py
@@ -9,7 +9,7 @@ import simplejson
 import requests
 
 from commcare_export.checkpoint import CheckpointManagerWithSince
-from commcare_export.commcare_hq_client import CommCareHqClient
+from commcare_export.commcare_hq_client import CommCareHqClient, ResourceRepeatException
 from commcare_export.commcare_minilinq import SimplePaginator, DatePaginator, resource_since_params, get_paginator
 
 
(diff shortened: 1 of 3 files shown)


tests/test_commcare_hq_client.py
```diff
@@ -102,6 +123,14 @@ def test_iterate_date(self):
         self._test_iterate(FakeDateFormSession(), get_paginator('form'), 3, [1, 2, 3])
         self._test_iterate(FakeDateCaseSession(), get_paginator('case'), 2, [1, 2])
 
+    def test_repeat_limit(self):
+        try:
```
czue, line 127: nit: it might more appropriate to use something like [`pytest.raises`](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest.raises) for this (offhand I couldn't remember what test stack we have)

solleks: Thanks, I've just pushed that change.

solleks:
It's a pleasure to have a round trip time shorter than one day. However, I will be going offline soon.

czue:
hmm it appears this broke py 2.7 tests. sorry about the hassle!

solleks:
Not too hard to fix. I will go ahead and merge.

PR #150, opened 2020-06-14
Add test for internal database errors

solleks:
This change is a short-term feature request to make database errors, including field size errors, clearer. It just catches SQLAchemy exceptions and prints them without any encumbering stack traces and side-effect exceptions.

Associated doc: https://docs.google.com/document/d/1YnS9mN2gxG-OBREoxLNgpWwIJJgIFW-4NU2QznKG0XM/edit?ts=5ede8253#


solleks:
Thanks! That takes care of the short-term feature requests.

PR #152, opened 2020-06-20
Pass data_types to TableSpec

solleks:
This change makes SQL table writers create the columns that have data types even when they don't have values.

I took out an assertion about column types because it proved to be very fragile. See the Travis CI test failure for the first commit for an example.


tests/test_cli.py
```diff
@@ -444,3 +444,55 @@ def test_cli_database_error(self, strict_writer, all_db_checkpoint_manager, capf
...
+        # set this so that it get's written to the checkpoints
+        checkpoint_manager.query = args.query
+
+        api_client_patch = mock.patch('commcare_export.cli._get_api_client',
+                                      return_value=DATA_TYPES_CLIENT)
+        # have to mock these to override the pool class otherwise they hold the db connection open
+        strict_writer_patch = mock.patch('commcare_export.cli._get_writer',
+                                         return_value=strict_writer)
+        checkpoint_patch = mock.patch('commcare_export.cli._get_checkpoint_manager',
+                                      return_value=all_db_checkpoint_manager)
```
czue, line 478: did you have to figure all these mocks out or were they duplicated from another test (if the latter it might be good to try and unify them)?

solleks: Good point. Since a lot of the code was like the _pull_data function, I factored it out into a new _pull_mock_data function.

PR #153, opened 2020-06-26
don't modify params in place

czue:
https://dimagi-dev.atlassian.net/browse/SAAS-11041

Introduced in https://github.com/dimagi/commcare-export/pull/149/

Because we are modifying the `params` object in place the [equality check](https://github.com/dimagi/commcare-export/pull/149/files#diff-e9b8e26e329e2f147322d44dd1799109R133) was always returning `True`, causing the request to fail after 10 tries.

### DIFF
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals, print_function, absolute_import, division, generators, nested_scopes
 
+import copy
 import logging
 from collections import OrderedDict
 
@@ -138,7 +139,7 @@ class CommCareHqClient(object):
                     raise ResourceRepeatException("Requested resource '{}' {} times with same parameters".format(resource, repeat_counter))
 
                 batch = self.get(resource, params)
-                last_params = params
+                last_params = copy.copy(params)
                 if not total_count or total_count == 'unknown' or fetched >= total_count:
                     total_count = int(batch['meta']['total_count']) if batch['meta']['total_count'] else 'unknown'
                     fetched = 0


commcare_export/commcare_hq_client.py
```diff
@@ -138,7 +139,7 @@ def iterate_resource(resource=resource, params=params):
                     raise ResourceRepeatException("Requested resource '{}' {} times with same parameters".format(resource, repeat_counter))
 
                 batch = self.get(resource, params)
-                last_params = params
+                last_params = copy.copy(params)
```
ShubhamBansal1997, line 142: don't we use deepcopy?

czue: i guess it'd be safer, but the object contains url parameters which are both not nested and have primitive values so i don't think it's necessary.

ShubhamBansal1997: cool, approving it

czue:
@solleks all good! thankfully we only know of one internal user who has run into it so far. :pray:

commcare_export/commcare_hq_client.py
millerdev: Where is `params` mutated? I don't see it.

nit: `dict` has a copy method
```suggestion
                last_params = params.copy()
```

solleks: It turns out that the return value of paginator.next_page_params_from_batch is a member variable called payload, so after assigning the return value once to params and then assigning to last_params, those two variables would remain aliases to the same dictionary.

czue: @millerdev I'm not 100% convinced that `params` will ever not be `None` so refraining form making that change for now. @solleks feel free to update / clean up as you see fit. My priority was getting a release out that did not have this bug in it as quickly as possible.

PR #154, opened 2020-07-10
support "numeric" as a valid data type

czue:
this allows you to use decimal values - one use case is for lat/lon from a GPS coordinate.

### DIFF
--- a/commcare_export/data_types.py
+++ b/commcare_export/data_types.py
@@ -8,6 +8,7 @@ DATA_TYPE_INTEGER = 'integer'
 DATA_TYPE_NUMERIC = 'numeric'
 
 DATA_TYPES_TO_SQLALCHEMY_TYPES = {
+    DATA_TYPE_TEXT: sqlalchemy.Text(),
     DATA_TYPE_BOOLEAN: sqlalchemy.Boolean(),
     DATA_TYPE_DATETIME: sqlalchemy.DateTime(),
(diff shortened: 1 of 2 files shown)


czue:
ugh looks like MS and MySQL both treat `Numeric` types as integers. wondering whether I should just change the test or figure out a way to make this smarter.

commcare_export/data_types.py
```diff
@@ -5,12 +5,15 @@
...
 DATA_TYPE_INTEGER = 'integer'
+DATA_TYPE_NUMERIC = 'numeric'
 
 DATA_TYPES_TO_SQLALCHEMY_TYPES = {
+    DATA_TYPE_TEXT: sqlalchemy.Text(),
     DATA_TYPE_BOOLEAN: sqlalchemy.Boolean(),
     DATA_TYPE_DATETIME: sqlalchemy.DateTime(),
     DATA_TYPE_DATE: sqlalchemy.Date(),
     DATA_TYPE_INTEGER: sqlalchemy.Integer(),
+    DATA_TYPE_NUMERIC: sqlalchemy.Numeric(),
```
snopoke, line 16: What if you use `DECIMAL`?

czue: Oh, good idea, I'll try that

czue: hmm, still looks like it's getting rounded. Will do a little more digging when I get time to set up MS or MySQL

PR #158, opened 2020-09-17
hex2uuid formats UUIDs

kaapstorm:
This allows UUIDs to be recognized by databases with a UUID data type, like PostgreSQL and SQL Server.

cc @czue

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ List of builtin functions:
 | str2date                     | Convert string to date                                                         |                                  |
 | bool2int                     | Convert boolean to integer (0, 1)                                              |                                  |
 | str2num                      | Parse string as a number                                                       |                                  |
-| hex2uuid                     | Parse a hex UUID, and format it into hyphen-separated groups                   |                                  |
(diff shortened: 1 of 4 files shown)


snopoke (commented):
This will still save it as a string in the DB. Is that OK or are you wanting to change the DB column type as well?

kaapstorm:
Yes, setting the value to a string is fine. Postgres and SQL Server accept string UUIDs for their UUID column types.

It seemed to me that for DET to set/change the column type, I'd need to create a SQLAlchemy UUID data type. For the use case I'm looking at, the column type has been set manually, so I'd like to use `--strict-types` so that DET does not try to change the column type back to TEXT.

Do you think the name "hex2uuid" is confusing, and people will expect it to set/change the column type to UUID? Or is it OK for people to set the column type manually and use the `--strict-types` argument? How do we generally feel about using `--strict-types`?

snopoke:
> Do you think the name "hex2uuid" is confusing, and people will expect it to set/change the column type to UUID?

I do think the name doesn't match the other XtoY functions. I think just renaming to `format-uuid` would be fine.

snopoke (approved):

PR #159, opened 2020-09-27
Use inserted_at to determine which forms/cases need updating

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-277

A client is reporting that occasionally properties aren't syncing correctly. @calellowitz helpfully came up with a theory that this is due to ES lag:

1) case updated at 9:27
2) DET run at 9:30 (causes checkpoint at 9:30)
3) case updated in ES at 9:32 (date modified is still 9:27)
4) next DET run checks only for cases modified after 9:30

The client is running the DET frequently, every 5 minutes.

This PR changes the date filtering to use `inserted_at`, the time the pillow inserted the item, instead of the server modified time. I don't think this will cause the exported last modified date to change, since it only updates the filters for getting forms/cases and then the pagination.

This will cause all case/forms to be resynced if the case or xform mappings change, although those mappings are quite stable. I'd guess it'll also cause resyncing when ES is upgraded, which might be a performance problem for us and/or for clients, although planning for that that seems better than living with the existing bug.


commcare_export/commcare_minilinq.py
```diff
@@ -41,7 +41,7 @@ def __call__(self, since, until):
             range_expression['lte'] = until.isoformat()
 
         server_modified_missing = {"missing": {
-            "field": "server_modified_on", "null_value": True, "existence": True}
+            "field": "inserted_at", "null_value": True, "existence": True}
```
snopoke, line 44: this complex filter is not necessary for `inserted_at` since all the records should have this data. It was only necessary for the modified on field since that was added without a full reindex so it was missing in older data.

See https://github.com/dimagi/commcare-export/pull/124/files#diff-5186f4dc371d1d524367519e15e4eae5L34

snopoke (changes requested):
@orangejenny ES lag should only be an issue if one Kafka partition is ahead of the another. The DET uses the modification time of the last seen document as the filter value for the next batch and not the 'time of run'.

Either way I think this is a good change if we are OK with the reindexing issue.

Also FYI:

* Issue [Switch to using date added to ES](https://github.com/dimagi/commcare-export/issues/107)
* [Previous PR for this issue](https://github.com/dimagi/commcare-export/pull/159)

commcare_export/commcare_minilinq.py
orangejenny: thanks, updated in 39c06df

orangejenny:
> ES lag should only be an issue if one Kafka partition is ahead of the another.

@snopoke How often does that happen? Do you think that makes it less plausible that this change will address the reported issue?

snopoke:
Looking at Datadog it seems to happen often enough that it could certainly be an issue: https://app.datadoghq.com/notebook/278293/change-lag-by-kafka-partition-case-form-pillows

orangejenny:
@snopoke Thank you! There's one specific example from the client, which is a case that the client's sql db says was last modified `2020-09-20 17:41`, but the real last modified date, from the API, is `2020-09-23 17:49`. Moving the notebook's date range to the afternoon of the 23rd, there's a 2.5-hour lag for cases starting around 18:15, so it does seem plausible that lag caused that case update to get missed.

snopoke:
I think I linked the wrong PR in my previous comment. Here's the correct link: https://github.com/dimagi/commcare-export/pull/124/files#diff-5186f4dc371d1d524367519e15e4eae5L34

commcare_export/commcare_minilinq.py
snopoke: I think the whole `FormFilterSinceParams` class can be removed and we can just use the `SimpleSinceParams` class (unless you think there is some ES data that doesn't have the `inserted_at` field.

orangejenny: makes sense, updated in ab9a9f4

commcare_export/commcare_minilinq.py
```diff
@@ -86,8 +43,8 @@ def __call__(self, since, until):
 
 def get_paginator(resource, page_size=1000):
     return {
-        'form': DatePaginator('form', ['server_modified_on','received_on'], page_size),
-        'case': DatePaginator('case', 'server_date_modified', page_size),
+        'form': DatePaginator('form', ['inserted_at','received_on'], page_size),
```
snopoke, line 46: Sorry I only noticed this now but I don't think we need to sort by both `inserted_at` and also `received_on`.

snopoke (changes requested):
Build still failing

commcare_export/commcare_minilinq.py
orangejenny: thanks, fixed in 7f03f7450ca02ea51433843035559fb840a493c3

orangejenny:
still working out some test failures

sravfeyn:
> I'd guess it'll also cause resyncing when ES is upgraded, which might be a performance problem for us and/or for clients

Could you please explain why upgrading should result in resyncing? @orangejenny (cc @millerdev )

orangejenny:
@sravfeyn I have no idea, I was just guessing it might.

snopoke:
If the upgrade does a reindex that's external ES then it could change the `indexed_on` field but the current plan is to do the reindex from the ES data itself in which case the date won't change

proteusvacuum:
Do we need a migration plan for people on older versions of the DET to start using this new pagination scheme? i.e. How will users  with old checkpoints based on `server_modified` be able to switch to using these new checkpoints? 
Maybe just saying "an upgrade to version X will require you to re-fetch all your data", although that doesn't seem like particularly friendly behaviour.

proteusvacuum:
@snopoke I just saw https://github.com/dimagi/commcare-export/pull/124#issuecomment-510038596 . Do you still think it is worth sorting by both `indexed_on` and `received_on`? I'm not sure I've fully wrapped my head around what that means for the user.

snopoke:
@proteusvacuum I don't think sorting on both is necessary.

The rollout for this is a bit tricky since we are switching the date fields. I think ideally we could create a migration that would right a migration that would re-load the last model that was exported, get the new date and update the checkpoint. Unfortunately we don't keep the ID of the last doc for the checkpoint (maybe something to add in the future).

tobiasmcnulty:
Hi all, I just wanted to chime in to say that we (and our client) are eagerly awaiting this fix. I'm happy to try it out if that's possible to do without breaking things irreparably. In any event, thanks for putting together this fix, and please let me know if there's anything we can do to help!

tobiasmcnulty:
P.S. In case you need some additional test data, I have ~516 cases that are refusing to update. Normally a duplicate bulk upload will fix the issue, but not for these cases. @orangejenny and @calellowitz know where to find me. 🙂

tobiasmcnulty:
Hi everyone. Our client is still experiencing this bug. Is there anything we can do from our side to help get this merged and released? We have the time and would like to help if we can. Many thanks, Tobias

czue:
Can we sidestep the migration issue by making the sort behavior configurable? I realize that increases the complexity of the API and maintenance burden, but perhaps having a seamless migration would be worth it? We could do something like:

Next release: add new sort param as an option
Following release: make new sort the default and deprecate old version
Release after that: remove old version

Looking at the code it'd be some work to introduce the option, but it looks manageable (especially if we've got @tobiasmcnulty help to get the work done).

@snopoke @orangejenny @proteusvacuum thoughts?

snopoke:
@czue that seems like a good workaround to me

orangejenny:
@czue I think that's a good idea

tobiasmcnulty:
Thanks everyone! @dpoirier on our team is looking into this and may have questions for you at some point. One point of clarification for me: Do you anticipate any server-side changes to support the migration, or is this all client-side?

czue:
@tobiasmcnulty I think the server-side changes are already rolled out in https://github.com/dimagi/commcare-hq/pull/24741

I think the basic process for @dporier would be to:

1. Add a CLI option to control the sort behavior.
2. Update [this code](https://github.com/dimagi/commcare-export/pull/159/files#diff-cc004dcac6b484cca9643fd2f51a0054593a32323abe3d69f5e502e7c791bbb3R34-R35) to make sure it honors the new CLI option, defaulting to the current behavior.
3. Add / update tests as necessary.

It's not clear to me whether it'd be easier to work in this PR versus start a new one. The additions made here are a good roadmap for how the new parameter should behave. The deleted code is how the existing parameter should behave and likely needs to be restored.

Hope that helps!

czue:
@tobiasmcnulty I'll just add that reading through the history, it's not clear to me that this change would solve this problem:

> P.S. In case you need some additional test data, I have ~516 cases that are refusing to update. Normally a duplicate bulk upload will fix the issue, but not for these cases. @orangejenny and @calellowitz know where to find me. slightly_smiling_face

A bulk upload should theoretically also update `server_modified_on` - unless we're doing something weird/smart to detect that you aren't making a meaningful update. So may also be worth digging in deeper into why that's happening.

dpoirier:
After looking at the PR & discussion and a bit of the code, I think that I would need a thorough knowledge of the data flows in CommCare in order to understand the right solution here, and I don't have that knowledge.

I don't know if this will be helpful, but I can share how we approached a synchronization problem on another project.

We added a "last_mod_time" field to the records, and used Django signals to ensure that the last_mod_time got updated whenever a record was changed, *or* whenever its many-to-many relationship to another model changed, that would also affect the synchronization. I would imagine database triggers could be used instead.

We tracked the last time each record had been synchronized, and when we next synchronized a record, we'd set the last sync time to the record's last mod time. (We tracked the last sync time in another table so we didn't have to worry about changes to the last sync time triggering another update to the last mod time, which would trigger another sync.)

When we wanted to synchronize records, we'd look for ones where the last sync time was not yet set, or the last sync time was before the last mod time. Because our synchronization was relatively slow (it'd take 4-5 days to synchronize all the records from scratch), we limited the number of records we'd sync each time, giving priority to the ones most recently modified. Our last mod time reflects user activity on a site, so synchronizing the most recently modified first means our most active users are most up-to-date, synchronization-wise. 

Again, I don't know how much of this would even be applicable to the CommCare case.

czue:
@dpoirier I don't think we need to come up with any fancy new synchronization schemes or anything. And I don't think the requested change requires any specific CommCare knowledge

This PR currently changes the sort order we request from CommCare HQ from one field to another.

The ask is to make that change configurable, instead of directly in the code. So basically just adding a flag that lets you toggle between the current behavior and the behavior that is attempting to be introduced here.

snopoke:
@dpoirier Are you likely to get to this in the next few days? Otherwise I can pick it up on Monday.

tobiasmcnulty:
Thanks for the follow up! A few questions/comments inline:

> @dpoirier I don't think we need to come up with any fancy new synchronization schemes or anything. And I don't think the requested change requires any specific CommCare knowledge

@czue I think what Dan was saying (this is true for me at least) is that it sounds like the issue _might_ have something to do with two different significantly complex services, ES and Kafka, and how data is synced between them on the backend. We're not familiar with how this is configured, so we're having a hard time judging if/how the proposed change solves the bug we're experiencing (or as you've suggested, it might not solve it at all).

Additionally, if the changes are all client-side, I'm unclear on how making the sort behavior configurable solves the rollout challenge. It might make more sense if I understood the proposed user instructions for going through this migration. Is the suggestion that they will need to switch back and forth between sort columns a couple times to make sure that everything is captured? Given that the bug has to do with missing data, personally I am not confident that switching sort columns would catch missed historical cases, and I would want to run a sync with `--start-over` one more time _after_ upgrading...

Also happy to discuss further here or offline! I am in our Caktus/Dimagi group on Skype and will ping you there.

> @dpoirier Are you likely to get to this in the next few days? Otherwise I can pick it up on Monday.

@snopoke Dan is out today, so I would say go ahead with the change if you have the bandwidth! Thanks.

snopoke:
> Thanks for the follow up! A few questions/comments inline:
> 
> > @dpoirier I don't think we need to come up with any fancy new synchronization schemes or anything. And I don't think the requested change requires any specific CommCare knowledge
> 
> @czue I think what Dan was saying (this is true for me at least) is that it sounds like the issue _might_ have something to do with two different significantly complex services, ES and Kafka, and how data is synced between them on the backend. We're not familiar with how this is configured, so we're having a hard time judging if/how the proposed change solves the bug we're experiencing (or as you've suggested, it might not solve it at all).
> 
> Additionally, if the changes are all client-side, I'm unclear on how making the sort behavior configurable solves the rollout challenge. It might make more sense if I understood the proposed user instructions for going through this migration. Is the suggestion that they will need to switch back and forth between sort columns a couple times to make sure that everything is captured? Given that the bug has to do with missing data, personally I am not confident that switching sort columns would catch missed historical cases. I think I would want to run a sync with `--start-over` one more time _after_ upgrading. What do you think?

Addressed offline but essentially the issue we're trying to mitigate is some portion of data being delayed before being accessible via the API. Since we currently use the modification date of the data to sort and checkpoint on this could result in missing data. Switching to 'indexed_on' guarantees that we will get data in the order it becomes available via the API.

The staged rollout is purely to allow us to get this feature out without forcing everyone to re-sync immediately.

@czue on that point, what do you think of using the old sorting for any existing tables and automatically using the new sorting for any new table as well as any time you pass in `--start-over`? That would allow us to roll it out in one go though it does make it a bit less visible to users.

czue:
> @czue on that point, what do you think of using the old sorting for any existing tables and automatically using the new sorting for any new table as well as any time you pass in --start-over? That would allow us to roll it out in one go though it does make it a bit less visible to users.

This seems like a great idea to me. I can't imagine any scenario where users would want visibility into this weird internal behavior, so hiding it seems like better overall UX.

snopoke:
@czue @orangejenny this is ready for review. Changes as follows:

* make pagination mode configurable within code
* store pagination mode in the checkpoint
* decide at start up which mode to use based on command line args and previous checkpoints
  * existing tables will continue with the previous pagination mode unless `--start-over` or `--since` are used
  * new tables will always use the new mode

tests/test_cli.py
```diff
@@ -421,7 +429,7 @@ class MockCheckpointingClient(CommCareHqClient):
     to return mocked data.
 
     Note this client needs to be re-initialized after use."""
-    def __init__(self, mock_data):
+    def     __init__(self, mock_data):
```
czue, line 432: was this intentional?

tests/test_cli.py
```diff
@@ -437,7 +445,11 @@ def __init__(self, mock_data):
     def get(self, resource, params=None):
         mock_requests = self.mock_data[resource]
         key = _params_to_url(params)
-        objects = mock_requests.pop(key)
+        try:
+            objects = mock_requests.pop(key)
+        except KeyError:
+            print(mock_requests.keys())
```
czue, line 451: nit: if this is intentional could consider adding a comment explaining why it's being printed out. print statements usually trigger a warning in my brain.

tests/test_cli.py
```diff
@@ -558,8 +576,100 @@ def test_cli_data_types_add_columns(self, strict_writer, all_db_checkpoint_manag
...
+        checkpoint_manager = all_db_checkpoint_manager.for_dataset("case", ["Case"])
+        # simulate previous run with legacy pagination mode
+        checkpoint_manager.set_checkpoint('2012-04-24T05:13:01', PaginationMode.date_modified, is_final=True)
+
+        _pull_mock_data(writer, all_db_checkpoint_manager, get_indexed_on_client(0), 'tests/013_ConflictingTypes.xlsx', start_over=True)
+        self._check_data(writer, [["doc 1"], ["doc 2"]], "Case")
+        self._check_checkpoint(checkpoint_manager, '2012-04-24T05:13:01', 'doc 2')
+
+    def test_cli_pagination_since(self, writer, all_db_checkpoint_manager):
+        """Test that we use to the new pagination mode when using 'since'"""
```
czue, line 659: this confused for a moment since it seems odd that you'd ever have a checkpoint and then manually override it. what's the use case for that?

czue (commented):
Wow, this looks great.

Caveats: I only reviewed code from https://github.com/dimagi/commcare-export/pull/159/commits/0f76ab90124f0e5b8048b9083b2b3667eb474245, and some of the changes to the tests made my head hurt a bit and I didn't fully take the time to understand them (but mostly made sense/looked good).

No need to change this - but I was wondering if we can safely flip from legacy to `indexed_on` in a regular checkpointed update. I think you would still sort/filter by the old value, but then theoretically you could use the `indexed_on` of the last doc to start the next round. There's probably an edge case in there that doesn't make it perfect (two docs, one has higher modified on and the other has higher indexed on that happen to fall right at the barrier?). You could also use an earlier indexed_on, but then you run the risk of doing extra work the next time around... :man_shrugging:

tests/test_cli.py
czue: (I think it got fixed in a later commit also)

czue (approved):
Oops, sorry meant to approve.

proteusvacuum (approved):
Very nice! Thanks for pushing this past the finish line.

tests/test_cli.py
snopoke: not sure if there is a use case but it's not disallowed so just wanted to test it. I also realized now that if you pass in `--since` or `--until` then we don't do checkpointing at all. I'm going to update the test to reflect that.

tests/test_cli.py
snopoke: not intentional, I'll remove it. Thanks

tests/test_cli.py
snopoke: will fix

commcare_export/commcare_minilinq.py
```diff
@@ -30,24 +41,64 @@ def __call__(self, since, until):
         return params
 
 
-SUPPORTED_RESOURCES = {
-    'form', 'case', 'user', 'location', 'application', 'web-user'
-}
+class FormFilterSinceParams(object):
```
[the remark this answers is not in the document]
snopoke: yes

snopoke:
> No need to change this - but I was wondering if we can safely flip from legacy to `indexed_on` in a regular checkpointed update. I think you would still sort/filter by the old value, but then theoretically you could use the `indexed_on` of the last doc to start the next round. There's probably an edge case in there that doesn't make it perfect (two docs, one has higher modified on and the other has higher indexed on that happen to fall right at the barrier?). You could also use an earlier indexed_on, but then you run the risk of doing extra work the next time around...

This is what I was suggesting here: https://github.com/dimagi/commcare-export/pull/159#issuecomment-717936452 but it's not straight forward to get the ID of the last doc from tables since they may not have a suitable column to sort on and the ID may also not be just the doc ID (in the case of exporting form repeats or cases from forms). We'd also then need to look up that doc in the API to get the `indexed_on` date. It's not impossible but I think it would at minimum require an intermediate release to ensure that the doc ID and date column are present in all tables.

tests/test_cli.py
snopoke: eb23b7f

tests/test_cli.py
snopoke: 8f57ed9

tests/test_cli.py
snopoke: 67f0624

snopoke (approved):

czue (approved):
thanks for follow ups!

czue:
:confetti_ball:

mikecjohn (commented):
The link in the log warning message to the Wiki does not exist, where did you want it to go? It currently points to: "https://wiki.commcarehq.org/display/commcarepublic/CommCare+Export+Tool+Release+Notes"

tobiasmcnulty:
The wiki and pypi all seem to refer back to GitHub Releases for release notes:
* https://confluence.dimagi.com/display/commcarepublic/CommCare+Data+Export+Tool
* https://pypi.org/project/commcare-export/
* https://github.com/dimagi/commcare-export/releases

czue:
Oops, thanks for the heads up! Confirming the last link is the correct one. https://github.com/dimagi/commcare-export/pull/177

PR #162, opened 2020-10-20
Strip column name whitespace

### LINKED ISSUE #161: Column Names with whitespace fail to parse
When column names in the excel query files have whitespace around them, these columns are no longer picked up by `commcare-export`, leading to very confusing results. We should ignore any whitespace around column names.

proteusvacuum:
Fixes #161


proteusvacuum:
I added some whitespace to a column name which is why that excel file is changed.

kaapstorm (approved):
Thank you!

PR #168, opened 2020-11-25
When writing to MS SQL, use special DATETIME2...

kmtracey:
...type for datetime objects. This avoids problems with the restricted
year range of default datetime type for MS SQL (1753 is oldest year
allowed for that type).

### DIFF
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -357,7 +357,10 @@ class SqlTableWriter(SqlMixin, TableWriter):
         if isinstance(val, bool):
             return sqlalchemy.Boolean()
         elif isinstance(val, datetime.datetime):
-            return sqlalchemy.DateTime()
+            if self.is_mssql:
+                return sqlalchemy.dialects.mssql.DATETIME2()
+            else:
+                return sqlalchemy.DateTime()
         elif isinstance(val, datetime.date):
             return sqlalchemy.Date()
 
--- a/tests/test_writers.py
+++ b/tests/test_writers.py
@@ -281,9 +281,8 @@ class TestSQLWriters(object):
             expected['bazzle']['c'] = '0'
         if 'pyodbc' in writer.connection.engine.driver:
             expected['bazzle']['c'] = '0'
-            # couldn't figure out how to make SQL Server convert date to ISO8601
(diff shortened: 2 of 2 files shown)


kmtracey:
Seems Travis build is going to fail on any external pr, per https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions. I did adjust one test that relied on the specifics of how a datetime value was converted to text in MSSQL. The remaining problem seem to be related to the encrypted values that can't be accessible in a PR from a fork.

snopoke (commented):
Will this convert existing columns to the new type? Is that conversion guaranteed to succeed?

kmtracey:
> Will this convert existing columns to the new type? Is that conversion guaranteed to succeed?

This does not change existing columns. (I'm not familiar enough with this code to know the answer based on the code, so I tried running an export with this code against a DB that had previously been partially populated without this change: existing datetime columns in a table in that DB remained datetime columns, they were not converted to datetime2 by the re-sync.)

It sounds like the datetime2 range is strictly larger than datetime, plus datetime2 has greater precision so I would think an auto-conversion would be OK in all cases but I don't really know enough MS SQL to be completely sure of that.

Not sure if it worth investigating/adding auto-conversion of existing columns here or if that would more efficiently be done by a recommendation to do it manually or drop/re-sync the database from scratch if the datetime2 type is wanted in the target of the sync?

mikecjohn:
Can we re-evaluate the suggested change in this PR? Even though this build failed, the recommendation should be considered. I think we should be specifying datetime2 for newer versions of MS SQL that use str2date conversion. Microsoft recommends this format for all "new work." 

https://docs.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver15

czue:
@snopoke this keeps coming up - do you know of any reason why we shouldn't merge this? (I can attempt to run tests locally or submit the PR from my account so we can ensure the build passes)

snopoke:
@czue I don't see any reason not to merge it assuming the tests pass.

czue (approved):

czue:
unfortunately, looks like the tests are still failing. https://travis-ci.com/github/dimagi/commcare-export/jobs/499034909 looks like it might be an easy fix and will put on my backlog to look unless @kmtracey you are available to help on this still?

snopoke:
Moved to https://github.com/dimagi/commcare-export/pull/219

PR #169, opened 2020-11-26
Set mssql nvarchar columns to fixed length to allow indexing

proteusvacuum:
We've had issues where columns are getting created in mssql that can't be indexed because the length is not getting set correctly.

https://dimagi-dev.atlassian.net/browse/SC-988?focusedCommentId=100916&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-100916

This change sets a max length, but then overrides this if any data is bigger than the 900 byte maximum. These resized columns will no longer be able to be indexed, but having indices on large columns seems funny anyway.

### DIFF
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -375,7 +375,13 @@ class SqlTableWriter(SqlMixin, TableWriter):
                 else:
                     return sqlalchemy.UnicodeText(collation=self.collation)
             elif self.is_mssql:
-                return sqlalchemy.NVARCHAR(collation=self.collation)
+                # MSSQL (pre 2016) doesn't allow indices on columns longer than 900 bytes
+                # - https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql
+                # If any of our data is bigger than this, then set the column to NVARCHAR(max)
+                # `length` here is the size in bytes - https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.String.params.length
+                length_in_bytes = len(val.encode('utf-8'))
+                column_length_in_bytes = None if length_in_bytes > 900 else 900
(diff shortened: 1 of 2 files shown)


commcare_export/writers.py
```diff
@@ -375,7 +375,13 @@ def best_type_for(self, val):
...
                     return sqlalchemy.UnicodeText(collation=self.collation)
             elif self.is_mssql:
-                return sqlalchemy.NVARCHAR(collation=self.collation)
+                # MSSQL (pre 2016) doesn't allow indices on columns longer than 900 bytes
+                # - https://docs.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql
+                # If any of our data is bigger than this, then set the column to NVARCHAR(max)
+                # `length` here is the size in bytes - https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.String.params.length
+                length_in_bytes = len(val.encode('utf-8'))
+                column_length_in_bytes = None if length_in_bytes > 900 else 900
+                return sqlalchemy.NVARCHAR(length=column_length_in_bytes, collation=self.collation)
```
proteusvacuum, line 384: I believe this will resize columns of everyone's DB currently in the wild. Maybe I should check somewhere that we aren't _shrinking_ columns? Is that even problematic?

snopoke: Resizing up seems OK, would be good to add a test to make sure it doesn't resize down.

kaapstorm (approved):

commcare_export/writers.py
proteusvacuum: 853fe0f0dea6b21af5c37b8a2cc729b895d48067

snopoke (approved):

PR #142, opened 2020-05-27
Add explicit data type support

czue:
https://trello.com/c/2J8FTZLO/40-data-export-tool-column-population and the [linked google doc](https://docs.google.com/document/d/17T2kbH1sBUoqhJOhnwXPfEYuTdWA0tjB-bsJL557OlA/edit#heading=h.eg8j7yioaqam)

This is the first major work I've done on DET so don't have a huge degree of confidence in it.

Review by commit. Hopefully it's mostly clear, but basically this adds a "data type" column, which you can set, and if you do then any SQL table will create that column the first time it runs (as long as there is at least one record). I'm not sure this was the *best* way to implement this, but it was the *easiest*.

I also changed the `table` dictionary to a `TableSpec` class, which I think is an improvement but open to feedback on that. It didn't end up actually being that useful for the work I wanted to do.

PR #170, opened 2020-12-11
don't evaluate the generator

snopoke:
This change https://github.com/dimagi/commcare-export/commit/842ad486cd2ee633fc5b8baeb14f91d1b1f565e6 which came from this comment https://github.com/dimagi/commcare-export/pull/142#discussion_r431918276 results in placing ALL the export data in memory before writing any of it to the output.

To fix that I've remove the code that eagerly evaluates the generator and also remove the rows from the equality check and JSON output of the TableSpec class (they definitely don't belong in either).

This should also fix https://dimagi-dev.atlassian.net/browse/SAAS-11588

Thanks to @bderenzi for noticing the weirdness and working through the debugging with me.

### DIFF
--- a/commcare_export/minilinq.py
+++ b/commcare_export/minilinq.py
@@ -434,7 +434,7 @@ class Emit(MiniLinq):
         env.emit_table(TableSpec(
             name=self.table,
             headings=[heading.eval(env) for heading in self.headings],
-            rows=list(map(self.coerce_row, rows)),
+            rows=map(self.coerce_row, rows),
             data_types=[lit.v for lit in self.data_types]
         ))
 
--- a/commcare_export/specs.py
+++ b/commcare_export/specs.py
@@ -13,7 +13,6 @@ class TableSpec:
             isinstance(other, TableSpec)
             and other.name == self.name
             and other.headings == self.headings
-            and other.rows == self.rows
             and other.data_types == self.data_types
         )
 
@@ -21,6 +20,5 @@ class TableSpec:
         return {
             'name': self.name,
             'headings': self.headings,
-            'rows': self.rows,
             'data_types': self.data_types,
         }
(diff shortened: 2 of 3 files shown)


snopoke:
This could also cause data loss since the SQL checkpoints are still written after each batch however the actual rows are only written after the last batch.

proteusvacuum (approved):

mikecjohn:
This is awesome, thank you everyone. Will someone be able to translate what exactly is changing in this fix in a way that I can explain it to the partners and end-users?

millerdev:
It looks like there are no tests for the bug that is fixed by this PR. Would it be hard to add them?

Considering that I had no idea that I was introducing a critical bug with [this recommendation](https://github.com/dimagi/commcare-export/pull/142#discussion_r431918276), test coverage around that area would be especially great.

czue:
wow nice find (and apologies for my contribution to this issue). @proteusvacuum @calellowitz this may also at least partially explain/fix the weird behavior we observed with resource constraints on the machines that were doing a lot of exports.

czue:
> Will someone be able to translate what exactly is changing in this fix in a way that I can explain it to the partners and end-users?

@mikecjohn not sure if you got an answer but my best attempt (probably good for @snopoke to confirm I got it right):

Previously there was a bug by which all new data was fetched from CommCare before being written to the export database. While data was being fetched, "checkpoints" were also saved. This caused two primary problems:

1. Performance issues when there was a large volume of new data to fetch - because all of the the resulting data had to be stored in memory before being written to the export DB.
2. Scenarios where some data would be missing from the export DB after an unexpected error occurred in writing to the database. A checkpoint could be written prior to the data being written, and if the write failed then the next run of the export would pick up from where the checkpoint was, creating a "gap" in the exported data.

PR #171, opened 2020-12-14
test that emitted rows is a generator

snopoke:
https://github.com/dimagi/commcare-export/pull/170#issuecomment-743282024

+ a small amount of reformatting


millerdev (commented):
Would it be possible to write a test that verifies that checkpoints are not written before the corresponding data? The test should verify destination db state, and would have failed before https://github.com/dimagi/commcare-export/pull/170/commits/180a7bd120ef6f0f8f09410dae0dd92e7ef0224c was added.

snopoke:
@millerdev e2b2715

millerdev (approved):

PR #175, opened 2021-02-17
support JSON columns in Postgres

czue:
This adds support for Postgres JSON column types. This is particularly useful if you want to use the DET to generate a "data lake" by exporting the entire contents of e.g. the "form" property.

I was shocked how easy this was. Only one little quirk to chase down with type compatibility.

### DIFF
--- a/commcare_export/data_types.py
+++ b/commcare_export/data_types.py
@@ -5,12 +5,14 @@ DATA_TYPE_BOOLEAN = 'boolean'
 DATA_TYPE_DATE = 'date'
 DATA_TYPE_DATETIME = 'datetime'
 DATA_TYPE_INTEGER = 'integer'
+DATA_TYPE_JSON = 'json'
 
 DATA_TYPES_TO_SQLALCHEMY_TYPES = {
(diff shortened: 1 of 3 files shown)


snopoke (approved):
Confirm that this only works if you explicitly specify the data type in the query file.

czue:
> Confirm that this only works if you explicitly specify the data type in the query file.

Correct. Was going to mention that in docs. I guess we could add a command-line flag to do this automatically or something, but guessing explicit is fine for now?

Let me know if you think it's worth adding a test that it doesn't change behavior without the type

snopoke:
> > Confirm that this only works if you explicitly specify the data type in the query file.
> 
> Correct. Was going to mention that in docs. I guess we could add a command-line flag to do this automatically or something, but guessing explicit is fine for now?

Yes, I think this should be explicit rather than creating JSON columns for incorrect query files.

snopoke:
> Let me know if you think it's worth adding a test that it doesn't change behavior without the type

Don't think it's necessary

proteusvacuum:
This is great!

PR #176, opened 2021-02-17
simplify db type check

czue:


czue:
nevermind, i guess this was necessary.

PR #184, opened 2021-06-07
update date filtering params for messaging-event

snopoke:
Merge after https://github.com/dimagi/commcare-hq/pull/29869 is deployed


czue (approved):
Do you know what's going on with the tests? The log is very noisy but didn't look related.

snopoke:
> Do you know what's going on with the tests? The log is very noisy but didn't look related.

Perhaps a change in the pyodbc driver and how it convert the connection string to the DSN. I also found another issue related to slqalchemy upgrade: https://github.com/dimagi/commcare-export/pull/185

PR #188, opened 2021-06-24
remove additional auto-id field

snopoke:
I believe this 'hack' become unecessary after this change [e1eae02417dbaeaf34c407bef0afe9600a21e24a](https://github.com/dimagi/commcare-export/commit/e1eae02417dbaeaf34c407bef0afe9600a21e24a).

Unfortunately this change may break existing usages of the DET if the 'auto_id' functionality is being used. This would be limited to a reference to an `id` field other than the root ID: `$.id` (or just `id` if not referencing a nested object).

### DIFF
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -200,10 +200,6 @@ class JsonPathEnv(Env):
 
         def iter(jsonpath_expr=jsonpath_expr): # Capture closure
             for datum in jsonpath_expr.find(self.__bindings):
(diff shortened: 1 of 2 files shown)


snopoke:
Related weird functionality in jsonpath-rw: https://github.com/kennknowles/python-jsonpath-rw/pull/96

snopoke:
Closing this for now since merging it will potentially break existing exports.

PR #189, opened 2021-06-24
Ability to emit the root document if the desired sub-document does not exist

snopoke:
The focus of the changes here is to allow exporting data from the messaging event API even when there are no messages. Each event may have multiple messages so in order to get one row per message we need an expression like this:
```
messaging-event.messages.[*]
```

However if there are no messages, as is the case when the event errors, we still want to export one row for the event. To accomplish this a new command line option is added to the CLI:

```
--export-root-if-no-subdocument
    Use this when you are exporting a nested document e.g. form.form..case, messaging-event.messages.[*]
    And you want to have a record exported even if the nested document does not exist or is empty.
```

When this option is used in conjunction with a Data Source query like those given in the examples, the minilnq expression
is adjusted so that if the main expression returns no results, the root document is used instead:

```diff
- Reference('messages.[*]')
+ Apply(Reference('_or_raw'), Reference('messages.[*]'), Reference("$"))
```

The `_or_raw` function is similar to `or` but it does not fully unwrap the values.

### DIFF
--- a/commcare_export/excel_query.py
+++ b/commcare_export/excel_query.py
@@ -252,7 +252,7 @@ def compile_source(worksheet, value_or_root=False):
         return data_source, api_query, None
     else:
         if value_or_root:
-            # if the jsonpath doesn't not yield a value yield the root document
+            # if the jsonpath doesn't yield a value yield the root document
             expr = get_value_or_root_expression(data_source_jsonpath)
         else:
             expr = Reference(str(data_source_jsonpath))
--- a/commcare_export/excel_query.py
+++ b/commcare_export/excel_query.py
@@ -204,7 +204,7 @@ def split_leftmost(jsonpath_expr):
         return (jsonpath_expr, jsonpath.This())
 
 
-def compile_source(worksheet):
+def compile_source(worksheet, value_or_root=False):
     """
     Compiles just the part of the Excel Spreadsheet that
     indicates the API endpoint to hit along with optional filters
@@ -260,7 +260,21 @@ def compile_source(worksheet):
     if data_source_jsonpath is None or isinstance(data_source_jsonpath, jsonpath.This) or isinstance(data_source_jsonpath, jsonpath.Root):
         return data_source, api_query, None
     else:
-        return data_source, api_query, Reference(str(data_source_jsonpath))
+        if value_or_root:
+            # if the jsonpath doesn't not yield a value yield the root document
+            expr = get_value_or_root_expression(data_source_jsonpath)
+        else:
+            expr = Reference(str(data_source_jsonpath))
+        return data_source, api_query, expr
+
+
+def get_value_or_root_expression(value_expression):
+    """Return expression used when iterating over a nested document but also wanting
+    a record if the value expression returns an empty result."""
+    return Apply(
+        Reference('_or_raw'), Reference(str(value_expression)), Reference("$")
+    )
+
 
 # If the source is expected to provide a column, then require that it is
 # already present or can be added without conflicting with an existing
@@ -296,9 +310,9 @@ def require_column_in_sheet(sheet_name, data_source, table_name, output_headings
 
     return (headings, body)
 
-def parse_sheet(worksheet, mappings=None, column_enforcer=None):
+def parse_sheet(worksheet, mappings=None, column_enforcer=None, value_or_root=False):
     mappings = mappings or {}
-    data_source, source_expr, root_doc_expr = compile_source(worksheet)
+    data_source, source_expr, root_doc_expr = compile_source(worksheet, value_or_root)
 
     table_name_column = get_column_by_name(worksheet, 'table name')
     if table_name_column:
@@ -355,7 +369,7 @@ class SheetParts(namedtuple('SheetParts', 'name headings source body root_expr d
         ]
 
 
-def parse_workbook(workbook, column_enforcer=None):
+def parse_workbook(workbook, column_enforcer=None, value_or_root=False):
     """
     Returns a MiniLinq corresponding to the Excel configuration, which
     consists of the following sheets:
@@ -378,7 +392,7 @@ def parse_workbook(workbook, column_enforcer=None):
     parsed_sheets = []
     for sheet in emit_sheets:
         try:
-            sheet_parts = parse_sheet(workbook[sheet], mappings, column_enforcer)
+            sheet_parts = parse_sheet(workbook[sheet], mappings, column_enforcer, value_or_root)
         except Exception as e:
             msg = 'Ignoring sheet "{}": {}'.format(sheet, str(e))
             if logger.isEnabledFor(logging.DEBUG):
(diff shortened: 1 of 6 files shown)


tests/test_minilinq.py
```diff
@@ -88,6 +85,33 @@ def test_eval_auto_id_reference_nested(self):
...
+        ]
+        value_or_root = Apply(
+            Reference('_or_raw'), Reference(str('bar.[*]')), Reference("$")
+        )
+        flatmap = FlatMap(source=Literal(data), body=value_or_root)
+        mmap = Map(source=flatmap, body=List([
+            Reference("id"), Reference('baz'), Reference('$.id'), Reference('$.foo.id'), Reference('$.foo.name')
+        ]))
+        self.check_case(mmap.eval(env), [
+            ['1.bar.1.bar.[0]', 'a1', '1', '1.bid', 'zip'],
```
czue, line 106: `id` is `parent_id.key.child_id`? and then if no child id is present then it's `parent_id.key.parent_id.key[index]`?

(mostly just confused by the first entry in this list)

czue: Actually if I'm being honest I'm having a hard time wrapping my head around several of these test cases. I appreciate the density of the test, but I wonder if some comments or named test cases might help a bit? Or if you think it's clear and the problem is my own understanding of minilinq that's fair too.

tests/test_minilinq.py
```diff
@@ -88,6 +85,33 @@ def test_eval_auto_id_reference_nested(self):
...
+            ['1.bar.bazzer', 'a2', '1', '1.bid', 'zip'],
+            ['2', [], '2', '2.bid', 'zap'],
+            ['3.bar.3.bar.[0]', [], '3', '3.bid', 'mip'],
```
czue, line 109: (also find it odd how different the `id` is in these two examples. it's because the dict counts as an entry but an empty list does not?)

commcare_export/excel_query.py
```diff
@@ -260,7 +260,21 @@ def compile_source(worksheet):
     if data_source_jsonpath is None or isinstance(data_source_jsonpath, jsonpath.This) or isinstance(data_source_jsonpath, jsonpath.Root):
         return data_source, api_query, None
     else:
-        return data_source, api_query, Reference(str(data_source_jsonpath))
+        if value_or_root:
+            # if the jsonpath doesn't not yield a value yield the root document
```
czue, line 264: nit: "doesn't not" --> "doesn't" (for some reason github is not letting me add suggestions)

tests/test_minilinq.py
```diff
@@ -88,6 +86,31 @@ def test_eval_auto_id_reference_nested(self):
...
+            {"id": 1, "foo": {'id': 'bid', 'name': 'zip'}, "bar": [{'baz': 'a1'}, {'baz': 'a2', 'id': 'bazzer'}]},
+            {"id": 2, "foo": {'id': 'bid', 'name': 'zap'}, "bar": []},
+            {"id": 3, "foo": {'id': 'bid', 'name': 'mip'}, "bar": {}},
+            # {"id": 4, "foo": {'id': 'bid', 'name': 'map'}, "bar": None},  # fails with TypeError from jsonpath
+            {"id": 5, "foo": {'id': 'bid', 'name': 'mop'}},
+        ]
+        value_or_root = get_value_or_root_expression('bar.[*]')
+        flatmap = FlatMap(source=Literal(data), body=value_or_root)
+        mmap = Map(source=flatmap, body=List([
+            Reference("id"), Reference('baz'), Reference('$.id'), Reference('$.foo.id'), Reference('$.foo.name')
```
czue, line 103: would it be worth adding a test case where `baz` exists on the parent (and the child is empty)? Is the behavior there that it would use it or that it would remain empty if it has fallen back to the parent emit (my guess is the desired behavior is to be empty)

czue (commented):
Code looks good and feature makes sense. Just had some small suggestions on testing.

tests/test_minilinq.py
snopoke: your confusion is well deserved. There are some 'bugs' and oddities in commcare-export as well as jsonpath-rw that cause this. I pushed a test to master earlier in the day to 'document' the weirdness: https://github.com/dimagi/commcare-export/blob/master/tests/test_minilinq.py#L58

I am planning to replace jsonpath-rw with jsonpath-ng which has a bunch of bugfixes and improvements and will resolve this specific issue (for this case 'id' will become '3').

czue: Ah, thanks I missed that. Sounds like I can mostly ignore the first column in these then? If so then everything is much clearer.

czue: (could we remove the id column from this test if it's already tested elsewhere and not part of what's being tested here?)

tests/test_minilinq.py
snopoke: good question, it would actually take the value from the parent:
```python
data = {"id": 6, "foo": {'id': 'bid', 'name': 'mop'}, "baz": "root_bazz"}
# id, baz, $.id, $.foo.id, $.foo.name
result = ['6', 'root_bazz', '6', '6.bid', 'mop']
```

That definitely isn't great. I have no idea how to work around that. I might be able to hack it by setting a special var in the env if we're in the case where we're using the root doc and then change the JsonPathEnv to only allow 'root' expressions.

czue: Hmm - I guess it's not possible to emit something like `{'$': root}` instead of the root document itself? (not sure if that even makes sense)

I don't know that it's an important enough edge-case to worry about, though does seem worth at least documenting the "expected" behavior in a test

snopoke: Not loving this but.... 8ac159f

tests/test_minilinq.py
snopoke: https://github.com/dimagi/commcare-export/pull/190

tests/test_minilinq.py
snopoke: It is confusing and results from both a bug in jsonpath-rw and a bug in commcare-export:

* https://github.com/kennknowles/python-jsonpath-rw/pull/96
* https://github.com/dimagi/commcare-export/pull/188

I would like to fix these issues but it would results in breaking certain exports (any that are using the feature that allows you to specify the root doc expression in the `Data Source` column.

I'm going to think about how we could do the rollout. If you have thought's I'd be keen to know.

czue:
looks like the other test now also needs to be updated

tests/test_minilinq.py
czue: The way they break is that the IDs change, yeah? So if one were using SQL, and, e.g. a case was modified, then you could end up with duplicate rows for the children?

Are you worried about migrating exports that Dimagi maintains or breaking external usages? In general introducing breaking changes seems maybe acceptable so long as it's a major version bump and it's clearly documented (including how to fix - maybe in this case just by wiping your data and starting over?). Certainly the expectation would be that one should read release notes for any major library upgrade.

We could also add a warning statement in a point release to log when someone would be affected by the upgrade.

czue (approved):
Good by me once tests pass. Thanks for filling in the gaps in my knowledge!

tests/test_minilinq.py
snopoke: > The way they break is that the IDs change, yeah? So if one were using SQL, and, e.g. a case was modified, then you could end up with duplicate rows for the children?

That's correct.

> Are you worried about migrating exports that Dimagi maintains or breaking external usages? In general introducing breaking changes seems maybe acceptable so long as it's a major version bump and it's clearly documented (including how to fix - maybe in this case just by wiping your data and starting over?). Certainly the expectation would be that one should read release notes for any major library upgrade.

I'm don't know of any that Dimagi maintains which use this feature. But if Dimagi is maintaining it we could work around the issue. I'm mostly concerned about those 'in the wild'.

> We could also add a warning statement in a point release to log when someone would be affected by the upgrade.

Yes, that's also a good idea.

snopoke: > Actually if I'm being honest I'm having a hard time wrapping my head around several of these test cases. I appreciate the density of the test, but I wonder if some comments or named test cases might help a bit? Or if you think it's clear and the problem is my own understanding of minilinq that's fair too.

Refactored tests: 4d194d2 (much better, thanks for the prompt)

czue (approved):
thanks for test cleanup!

PR #190, opened 2021-06-25
switch to jsonpath-ng

snopoke:
PR on top of https://github.com/dimagi/commcare-export/pull/189

jsonpath-ng is a fork of jsonpath-rw with some additional features and a number of bugfixes. Since jsonpath-rw is not currently maintained it makes sense to move to jsonpath-ng. jsonpath-ng is also used in HQ.

### DIFF
--- a/commcare_export/env.py
+++ b/commcare_export/env.py
@@ -9,8 +9,8 @@ import pytz
 import six
 from itertools import chain
 
-from jsonpath_rw import jsonpath
-from jsonpath_rw.parser import parse as parse_jsonpath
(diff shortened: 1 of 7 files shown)


czue (approved):
code looks good for whenever you have a migration plan. are there not any other tests that need to change

snopoke:
> code looks good for whenever you have a migration plan. are there not any other tests that need to change?

I don't think any migration is necessary. This is basically a drop in replacement. Installing a new version of commcare-export will automatically install jsonpath-ng.

PR #193, opened 2021-08-06
prompt for username & password before they are needed

snopoke:
Get the username and password from the user before they are used to create the API client.

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -300,6 +300,13 @@ def main_with_args(args):
               '--query, --users, --locations', file=sys.stderr)
         return EXIT_STATUS_ERROR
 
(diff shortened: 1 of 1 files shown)


mkangia (approved):
Interesting. Thank you.

Just confirming if is from the ticket that was submitted for the issue where cli wasn't working with prompts because that was marked as done recently or you just fixed it?
Can you add description?

snopoke:
> Interesting. Thank you.
> 
> Just confirming if is from the ticket that was submitted for the issue where cli wasn't working with prompts because that was marked as done recently or you just fixed it?

Yes, it's from that ticket: https://dimagi-dev.atlassian.net/browse/SAAS-12490

PR #194, opened 2021-09-22
FIX: Alembic 1.7.x is breaking migrations

bderenzi:
I assume there's some step after this about cutting a new release of the DET, but not sure how to do that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setuptools.setup(
     include_package_data=True,
     license = 'MIT',
     install_requires = [
-        'alembic<=1.6.5',
+        'alembic<1.7',
         'argparse',
         'jsonpath-ng~=1.5',
         'openpyxl==2.5.12',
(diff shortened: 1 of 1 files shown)


czue:
FYI, the release process is documented here: https://github.com/dimagi/commcare-export/#release-process

It's scope creep for this PR, but perhaps we should also switch this to pip-tools / requirements files sometime. Guessing we should have pinned versions of all dependencies.

setup.py
```diff
@@ -67,7 +67,7 @@ def run_tests(self):
     include_package_data=True,
     license = 'MIT',
     install_requires = [
-        'alembic',
+        'alembic<=1.6.5',
```
millerdev, line 70: Since 1.7 is the breaking version, why not use that as the limit? That way newer patch release to 1.6 would still be allowed.
```suggestion
        'alembic<1.7',
```

snopoke (approved):

millerdev (approved):

PR #200, opened 2021-11-09
Update readme with pinned version of alembic

proteusvacuum:
New users installing commcare-export end up with an unsupported alembic version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ $ pip install openpyxl
 $ pip install xlwt
 
 # To sync with a SQL database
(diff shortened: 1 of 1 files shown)


czue (approved):

snopoke (approved):
It would be nice to setup the packaging so that these extras are defined in the package and installation can be done with a simpler command like:
```
pip install commcare-export[sql]
```

https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies

kaapstorm (approved):

PR #205, opened 2022-04-19
Black

kaapstorm:
Personally, I don't really like [Black](https://github.com/psf/black). I think a careful programmer can do a better job at applying a style guide and making good discretionary choices to optimize readability.

But the commcare-export codebase is old, and not all programmers have been equally, well, obsessive. So this is a clumsy shortcut to avoid the pain one gets behind the eyes when one encounters a line longer than 115 characters. (I don't know, is it just me?)

This PR doesn't change any functionality, although a few "%" and "format()" strings are replaced with f-strings.

### DIFF
--- a/black.toml
+++ b/black.toml
@@ -1,0 +1,4 @@
+[tool.black]
+target-version = ['py36', 'py37', 'py38']
+line-length = 79
+skip-string-normalization = true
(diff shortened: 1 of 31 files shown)


proteusvacuum:
On `commcare-hq` we use yapf: https://github.com/dimagi/commcare-hq/pull/29236 
Should we be consisten?

kaapstorm:
Ah! Yes, we should definitely be consistent and use that here too.

PR #207, opened 2022-04-22
Use YAPF instead of Black

kaapstorm:
Uses YAPF to clean up the code a bit. It's not perfect, but it's quick and it's better.

Context: https://github.com/dimagi/commcare-export/pull/205

### DIFF
--- a/commcare_export/builtin_queries.py
+++ b/commcare_export/builtin_queries.py
@@ -11,6 +11,7 @@ excel_query.blacklist(LOCATIONS_TABLE_NAME)
 
 
 class Column:
+
     def __init__(self, name, source, map_function=None, *extra_args):
         self.name = Literal(name)
         self.source = source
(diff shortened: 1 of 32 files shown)


snopoke (approved):
I didn't review all the changes in detail but I'm assuming it's all formatting according to the YAPF config. Please call out anything that doesn't fit into that assumption.

kaapstorm:
That's correct. No functional changes in here, just formatting, and a few "%" and "format()" strings written as f-strings.

commcare_export/builtin_queries.py
```diff
@@ -22,22 +23,27 @@ def mapped_source_field(self):
...
 
 def compile_query(columns, data_source, table_name):
-    source = Apply(Reference('api_data'), Literal(data_source),
-                   Reference('checkpoint_manager'))
-    part = excel_query.SheetParts(table_name, [c.name for c in columns], source,
-                                  List([c.mapped_source_field for c in columns]),
-                                  None)
+    source = Apply(
+        Reference('api_data'), Literal(data_source),
+        Reference('checkpoint_manager')
```
millerdev, line 35: This one is a bit annoying. Wonder why it didn't do
```suggestion
        Reference('api_data'),
        Literal(data_source),
        Reference('checkpoint_manager')
```

commcare_export/builtin_queries.py
```diff
@@ -22,22 +23,27 @@ def mapped_source_field(self):
...
+    )
+    part = excel_query.SheetParts(
+        table_name, [c.name for c in columns], source,
+        List([c.mapped_source_field for c in columns]), None
```
millerdev, line 39: Same here. Per my aesthetic preferences
```
this_style(is_for, cramming, as_many, args,
           on, each, line, as_is, possible)

# and

this_other_better_style(
    puts_each,
    argument,
    on_a,
    separate,
    line,
)
```
When I see the second, I expect each line to be a separate argument, so what YAPF is doing here is a little unexpected.

I forget if YAPF has an option for that?

millerdev (approved):
Also didn't review every line, but it seems like a big improvement. Thanks!

commcare_export/builtin_queries.py
kaapstorm: Yes! Great suggestion. See https://github.com/dimagi/commcare-export/pull/207/commits/f7ff584d5eb5640f734fa016c1fade20feb18a18

You have to be a bit careful, cos that setting can mess with comprehensions:
```python
    {foo: bar for foo, bar in some_dictionary.items() if do_i_want_this_one(bar)}
```
becomes
```python
    {foo: bar for foo,
     bar in some_dictionary.items() if do_i_want_this_one(bar)}
```

But if you unpack into a tuple, it does a better job:
```python
    {foo: bar for (foo, bar) in some_dictionary.items() if do_i_want_this_one(bar)}
```
becomes
```python
    {foo: bar for (foo, bar) in some_dictionary.items()
     if do_i_want_this_one(bar)}
```

PR #211, opened 2022-04-24
Move types from docstrings into type hints

kaapstorm:
Controversial, I know. But `env.py` and `minilinq.py` have had type hints since 2013. They were just in docstrings. This change converts what was written in the docstrings to actual type hints, and adds automated checking with Travis.

### DIFF
--- a/commcare_export/minilinq.py
+++ b/commcare_export/minilinq.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any
+from typing import Any, Dict
 from typing import List as ListType
 from typing import Optional
 
@@ -25,7 +25,7 @@ class MiniLinq(object):
 
     #### Factory methods ####
 
-    _node_classes = {}
+    _node_classes: Dict[str, 'MiniLinq'] = {}
 
     @classmethod
     def register(cls, clazz, slug=None):
@@ -71,6 +71,9 @@ class MiniLinq(object):
                 )
 
             return cls._node_classes[slug].from_jvalue(jvalue)
+
+    def to_jvalue(self):
+        raise NotImplementedError()
 
 
 class Reference(MiniLinq):
@@ -175,11 +178,6 @@ class Bind(MiniLinq):
                 'body': self.body.to_jvalue()
             }
         }
-
-    def __repr__(self):
-        return '%s(name=%r, value=%r, body=%r)' % (
-            self.__class__.__name__, self.name, self.value, self.body
-        )
 
 
 class Filter(MiniLinq):
@@ -476,8 +474,8 @@ class Emit(MiniLinq):
     def __init__(
         self,
         table: str,
-        headings: ListType[str],
-        source: ListType[MiniLinq],
+        headings: ListType[MiniLinq],
+        source: MiniLinq,
         missing_value=None,
         data_types=None,
     ):
--- a/commcare_export/minilinq.py
+++ b/commcare_export/minilinq.py
@@ -1,6 +1,9 @@
 import logging
+from typing import Any
 from typing import List as ListType
+from typing import Optional
 
+from commcare_export.env import Env
 from commcare_export.misc import unwrap, unwrap_val
 from commcare_export.repeatable_iterator import RepeatableIterator
 from commcare_export.specs import TableSpec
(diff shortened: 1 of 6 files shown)


kaapstorm:
(This change does not add type hints to all functions in `env` and `minilinq` -- only to the functions that had types in their docstrings. So it might not be perfect, but it feels like an improvement on the docstrings.)

mypy.ini
```diff
@@ -0,0 +1,35 @@
...
+implicit_reexport = False
+warn_unreachable = True
+
+# Reporting
+show_error_codes = True
+
+
+# Per-module options:
+
+# [mypy-commcare_export.minilinq]
```
proteusvacuum, line 34: nit: commented

commcare_export/utils.py
```diff
@@ -50,8 +51,9 @@ def print_runs(runs):
 
     StreamingMarkdownTableWriter(
         sys.stdout, compute_widths=True
-    ).write_table({
```
proteusvacuum, line 53: was this just not working before?

proteusvacuum (approved):
Approved, but I'm not clear on what our stance on type hints is exactly...

commcare_export/utils.py
kaapstorm: It was written in 2018, and I think the parameter type was changed from a dict to a TableSpec in 2020. Currently I get

```
$ python commcare_export/utils_cli.py history \
     --query example_det_config.xlsx \
     --project demo \
     --output postgresql://postgres:postgres@localhost:5432/postgres

Listing checkpoints (most recent 10):
    project:        demo
    query filename: example_det_config.xlsx

Traceback (most recent call last):
  File "commcare_export/utils_cli.py", line 167, in <module>
    entry_point()
  File "commcare_export/utils_cli.py", line 163, in entry_point
    main(sys.argv[1:])
  File "commcare_export/utils_cli.py", line 154, in main
    exit(main_with_args(args))
  File "commcare_export/utils_cli.py", line 159, in main_with_args
    command().run(args)
  File "commcare_export/utils_cli.py", line 62, in run
    print_runs(runs)
  File "commcare_export/utils.py", line 51, in print_runs
    StreamingMarkdownTableWriter(
  File "commcare_export/writers.py", line 233, in write_table
    col_widths = self._get_column_widths(table)
  File "commcare_export/writers.py", line 259, in _get_column_widths
    all_rows = [table.headings] + table.rows
AttributeError: 'dict' object has no attribute 'headings'

```

sravfeyn (approved):

commcare_export/utils.py
kaapstorm: Yeah, the jury is still out wrt to type hints in HQ. But the type hints in commcare-export have been here for 9 years. I just made them parseable, and fixed them. And then fixed the bug that the now-parseable type hints exposed.

If this seems like a step backwards, we can revert this PR -- it's why I moved the typing stuff into its own branch But to me it feels like I made some unhelpful docstrings a bit more helpful.

commcare_export/minilinq.py
```diff
@@ -459,8 +475,8 @@ class Emit(MiniLinq):
     def __init__(
         self,
         table: str,
-        headings: ListType[str],
-        source: ListType[MiniLinq],
+        headings: ListType[MiniLinq],
+        source: MiniLinq,
         missing_value=None,
         data_types=None,
```
snopoke, line 481: should these 2 also have hints?

millerdev:
> env.py and minilinq.py have had type hints since 2013.

Was the formal concept of type hints actually a thing back then? Calling these "type hints" as we know the concept today seems like a bit of a stretch to me. The [first type hints PEP](https://peps.python.org/pep-0484/) was only accepted for Python 3.5 in May of 2015. I don't believe this tool worked on Python 3 back then, but there was a different comment syntax `# type: (arg_type, ...) -> return_type` for Python 2.7 that was also outlined in PEP-484. It seems like the comment format chosen by the authors of those modules just happened to look like something similar to what type hints would eventually become.

snopoke:
> > env.py and minilinq.py have had type hints since 2013.
> 
> Was the formal concept of type hints actually a thing back then? Calling these "type hints" as we know the concept today seems like a bit of a stretch to me. The [first type hints PEP](https://peps.python.org/pep-0484/) was only accepted for Python 3.5 in May of 2015. I don't believe commcare-export worked on Python 3 back then, but there was a different comment syntax `# type: (arg_type, ...) -> return_type` for Python 2.7 that was also outlined in PEP-484. It seems like the comment format chosen by the authors of those modules just happened to look similar to what type hints would eventually become.

I think it's clear that the intention of the comments was to 'hint at the types', regardless of whether they matched any specific PEP.

commcare_export/minilinq.py
kaapstorm: https://github.com/dimagi/commcare-export/pull/213

PR #212, opened 2022-04-25
We don't need to reflect all the tables

kaapstorm:
Context: [SC-2022](https://dimagi-dev.atlassian.net/browse/SC-2022)

This change improves performance by avoiding [reflecting all the tables](https://docs.sqlalchemy.org/en/14/core/reflection.html#reflecting-all-tables-at-once) in a database.

This change has been tested with:
- [x] Adding a table with `--strict-types`
- [x] Adding a table without `--strict-types`
- [x] Adding columns to an existing table with `--strict-types`
- [x] Adding columns to an existing table without `--strict-types`

Easier to :blowfish: review :tropical_fish: by :fish: commit.

### DIFF
--- a/commcare_export/writers.py
+++ b/commcare_export/writers.py
@@ -510,10 +510,7 @@ class SqlTableWriter(SqlMixin, TableWriter):
             self._create_table(table_name, row_dict, data_type_dict)
             return
 
-        def get_current_table_columns():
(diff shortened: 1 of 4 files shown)


Charl1996 (approved):
Nice!

proteusvacuum:
When you are done with these changes, it will be worth cutting a new version.

PR #213, opened 2022-04-26
Add missing types

kaapstorm:
I accidentally left these out.

### DIFF
--- a/commcare_export/minilinq.py
+++ b/commcare_export/minilinq.py
@@ -477,9 +477,9 @@ class Emit(MiniLinq):
         table: str,
         headings: ListType[MiniLinq],
         source: MiniLinq,
-        missing_value=None,
-        data_types=None,
-    ):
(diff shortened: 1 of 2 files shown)


commcare_export/minilinq.py
```diff
@@ -477,9 +477,9 @@ def __init__(
         table: str,
         headings: ListType[MiniLinq],
         source: MiniLinq,
-        missing_value=None,
-        data_types=None,
-    ):
+        missing_value: Optional[str] = None,  # Denotes "no value". e.g. "---"
```
snopoke, line 480: style question... should this be here or in the docstring? I don't think I've seen per argument comments like this before.

snopoke (approved):

commcare_export/minilinq.py
kaapstorm: Great question. I was unsure.

I'll create a docstring for `__init__()`. Some of those other params could also use a bit of explaining.

PR #217, opened 2023-02-01
Preserve values of duplicate filter names

kaapstorm:
Context: [SC-2201](https://dimagi-dev.atlassian.net/browse/SC-2201)
PR for Case List API support: https://github.com/dimagi/commcare-hq/pull/32566

If a filter name is repeated with multiple filter values, commcare-export will pass both values to the API, not just the last one.

Once the PR linked above is merged, the Case List API will return cases that match any of the values. (In other words, an "OR" filter will be applied to multiple values of the same filter name.)

Tested locally.

fyi @joxl / SAAS

### DIFF
--- a/commcare_export/excel_query.py
+++ b/commcare_export/excel_query.py
@@ -122,7 +122,12 @@ def compile_filters(worksheet, mappings=None):
             for cell in get_column_by_name(worksheet, 'filter value') or []
         ]
     )
-    return zip(filter_names, filter_values)
+    # Preserve values of duplicate filter names. Results in an OR filter.
+    # e.g. {'type': ['person'], 'owner_id': ['abc123', 'def456']}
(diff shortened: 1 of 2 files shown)


calellowitz (approved):
Assuming you have tested it and confirmed the expected behavior, this looks good to me.

sravfeyn (approved):

kaapstorm:
Tests passed for the Python 3.6 build, but failed for Python 3.7 and 3.8 builds. This happened on [the last PR to be merged too](https://github.com/dimagi/commcare-export/pull/215) -- 3.6 succeeded,but 3.7 and 3.8 failed, but for a different reason than this time.

I'm going to merge, but I will create a ticket to update Travis: It should be building against Python 3.10 now.

PR #218, opened 2023-03-07
Add test.yml in .github/workflows

MartinRiese:
Context: [USH-1425](https://dimagi-dev.atlassian.net/browse/USH-1425)

* Most of the command were copied one to one from `travis.yml` to `.github/workflows/test.yml`
* Python 3.6 was not by the same ubuntu docker. Since it an older version I dropped it and added checks for 3.9 and 3.10 instead
* The builds for the new pythons versions failed originally since they pulled sqlalchemy 2 in. So I fixed the version to 1.4
* Coverage report shows up on coverall.io: https://coveralls.io/builds/57743106


setup.py
```diff
@@ -86,7 +86,7 @@ def run_tests(self):
         'pytz',
         'requests',
         'simplejson',
-        'sqlalchemy',
+        'sqlalchemy~=1.4',
```
MartinRiese, line 89: No sure if we would want a story to update the code to support version 2

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9, '3.10'] # Version 3.6 with arch x64 not found
```
MartinRiese, line 34: Added 3.9 and 3.10. Should we drop more than 3.6?

snopoke: I think we should keep it until it's past EOL (2023-06-27).

.github/workflows/test.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          fetch-depth: 50
+      - run: git fetch --tags origin # So we can use git describe. actions/checkout@v3 does not pull tags.
+
+      # MySQL set up
+      - run: sudo service mysql start # Ubuntu already includes mysql no need to use service
+      - run: mysql -uroot -proot -e "CREATE USER '${{ env.DB_USER }}'@'%';"
+      - run: mysql -uroot -proot -e "GRANT ALL PRIVILEGES ON *.* TO '${{ env.DB_USER }}'@'%';"
+
+      - uses: actions/setup-python@v3
```
snopoke, line 46:
```suggestion
      - uses: actions/setup-python@v4
```

Updating to v4 should fix the 'set-output' warnings

snopoke (approved):
One comment but not blocking.

PR #222, opened 2023-08-23
Add UCR support

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2911)

Background:
This work adds support to the DET for exporting data sources from CommCareHQ using the `api_get_ucr_data` API endpoint.

Due to the generic nature of the DET, not much work was needed to make the UCR exports compatible. 

One thing to note:
The datasources on HQ is stored in a custom table. This table does not have an `id` column and uses the `doc_id` column as the primary key. When HQ gives the datasource data it adds an `id` column (in addition to the `doc_id` column) which contains the same value as the `doc_id` column, but this new `id` column is the primary key in the DET database. If this is a problem I can change that, but that will require some additional changes to the DET.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,6 +48,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - run: sudo apt-get install pandoc
       - run: python setup.py sdist
+      - run: pip install --upgrade pip
       - run: pip install dist/*
       - run: pip install pymysql psycopg2 pyodbc
       - run: pip install coverage coveralls
--- a/commcare_export/commcare_minilinq.py
+++ b/commcare_export/commcare_minilinq.py
@@ -20,7 +20,8 @@ SUPPORTED_RESOURCES = {
     'location',
     'application',
     'web-user',
-    'messaging-event'
+    'messaging-event',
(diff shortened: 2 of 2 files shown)


snopoke (commented):
With this implementation the DET will not record any checkpoints (checkpoints currently only support the `DatePaginator`). This means that each run of the DET will attempt to re-export the full UCR table.

To support checkpoints I think we could do one of 2 things:
1. (preferred) Update the checkpoint to allow storing the 'cursor' value
2. Use our knowledge of the internals of the API create a new 'DateCursorPaginator' which extracts the date and doc ID from the cursor to store in the checkpoint.

Note that we don't currently use 'checkpoint.last_doc_id' at all. If we want to use it in pagination we'll need to make some other code changes to pass it (or the whole checkpoint) through.

Charl1996:
@snopoke Thanks for highlighting that (and sorry for only coming back to you now). 

I'm almost done adding the checkpoint support, after which I will create a PR. For now, though, I'm officially opening this PR for review.

snopoke (approved):
FYI, https://github.com/dimagi/commcare-export/pull/224 should fix the build

PR #223, opened 2023-08-30
Add checkpoint support for datasources

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-3091)

This PR adds checkpointing support for the datasources export runs by storing the `cursor` value returned from HQ to as an indication of progress. 

I tested this by verifying that the last stored `cursor` value in the `commcare_export_runs` table corresponds to the `cursor` value populated in the initial parameters when making the first request when the script reruns.


commcare_export/checkpoint.py
```diff
@@ -442,14 +448,19 @@ def __init__(
         self.since = since
         self.base_checkpoint_manager = base_checkpoint_manager
 
-    def get_since(self, checkpoint_manager):
+    def get_since(self, checkpoint_manager, data_source=None):
```
snopoke, line 451: What about using a new `PaginationMode` value instead of passing the data source here? Just thinking that in future there may be other 'cursor' based pagination APIs.

commcare_export/commcare_hq_client.py
```diff
@@ -232,6 +233,13 @@ def checkpoint(self, checkpoint_manager, paginator, batch, is_final):
                 logger.warning(
                     'Failed to get a checkpoint date from a batch of data.'
                 )
+        if isinstance(paginator, UCRPaginator):
+            cursor = paginator.next_page_params_from_batch(batch)['cursor'][0]
+            checkpoint_manager.set_checkpoint(
+                datetime.utcnow(),
+                is_final,
+                cursor=cursor,
+            )
```
snopoke, line 242: What do you think about pushing this to the paginator class?:

```
def set_checkpoint(checkpoint_manager, batch, is_final):
```

snopoke (commented):
Looking good.

commcare_export/commcare_hq_client.py
Charl1996: @snopoke For that matter the whole `checkpoint` method can be pushed to the respective paginator classes (`DatePaginator` or `UCRPaginator`)

Charl1996: @snopoke Something like [this](https://github.com/dimagi/commcare-export/pull/223/commits/90ba91789191a71ecb92bd1871933b41c94c01f9)?

Charl1996:
Not sure why the tests now fail. It works locally.

commcare_export/checkpoint.py
```diff
@@ -462,6 +472,9 @@ def get_pagination_mode(self, checkpoint_manager):
         if self.start_over or self.since or not checkpoint_manager:
             return PaginationMode.date_indexed
 
+        if checkpoint_manager.data_source == 'ucr':
```
snopoke, line 475: should this go above the previous `if` statement?

Charl1996: @snopoke I think technically you're right. I initially thought of doing that, but then tested it and it seemed fine, although I think my testing might not have included all the scenarios, so I'll move it.

snopoke:
Most test failures seem to be caused by this:

```
commcare_export/checkpoint.py:472: in get_pagination_mode
    if checkpoint_manager.data_source == 'ucr':
E   AttributeError: 'NoneType' object has no attribute 'data_source':
```

Charl1996:
Now it's the previous failures again.

snopoke:
Tests failing due to a change in `jsonpath-ng`. I've pushed 2 commits to fix it.

Charl1996:
@snopoke Goodness, I would have taken a long time to figure that out. Thanks!

snopoke (approved):

PR #225, opened 2023-10-20
Respect the Retry-After header

SmittieC:
- The DET will now sleep for the amount of seconds specified by the `Retry-After` header when it receives a 429 status from HQ.
- Increased the default page size that UCR's will be exported with. 
- Changed the way we set the default page size.

[HQ PR](https://github.com/dimagi/commcare-hq/pull/33572) to return the `Retry-After` header


snopoke (commented):
You should be able to use the backoff decorator to implement the 'retry-after' rather than doing it separately: https://github.com/litl/backoff#backoffruntime

commcare_export/commcare_minilinq.py
```diff
@@ -224,7 +225,8 @@ class DatePaginator(SimplePaginator):
 
     DEFAULT_PARAMS = object()
 
-    def __init__(self, since_field, page_size=1000, params=DEFAULT_PARAMS):
+    def __init__(self, since_field, page_size=None, params=DEFAULT_PARAMS):
+        page_size = page_size if page_size else 1000
```
zandre-eng, line 229: Is there a reason why the default value for `page_size` was removed and was instead put into this conditional check?

commcare_export/commcare_minilinq.py
```diff
@@ -177,7 +177,8 @@ class SimplePaginator(object):
     Paginate based on the 'next' URL provided in the API response.
     """
 
-    def __init__(self, page_size=1000, params=None):
+    def __init__(self, page_size=None, params=None):
+        page_size = page_size if page_size else 1000
```
zandre-eng, line 181: Nit: Could we put 1000 into a const variable? Perhaps something like `DEFAULT_PAGE_SIZE`

zandre-eng (approved):
Looks good, just a few non-blocking comments.

commcare_export/commcare_minilinq.py
SmittieC: https://github.com/dimagi/commcare-export/pull/225/commits/a32d003bcf74eda938b8aa869801870b0b2685b5

commcare_export/commcare_minilinq.py
zandre-eng: Ah, interesting! I never knew about this "gotcha".

SmittieC:
@snopoke are you happy with this PR?

snopoke (approved):

commcare_export/commcare_hq_client.py
```diff
@@ -30,6 +32,12 @@
 
 def on_backoff(details):
     _log_backoff(details, 'Waiting for retry.')
+    response = details["exception"].response
+    if response.status_code == 429:
+        retry_after = response.headers.get("Retry-After", 0.0)
+        retry_after = ceil(float(retry_after))
+        logger.warning(f"Sleeping for {retry_after} seconds")
+        time.sleep(retry_after)
```
snopoke, line 40: this isn't what I had in mind but I guess it works.

SmittieC: Another way is to use the `wait_gen` parameter to give it the time to wait for, but it would need to be context aware. Currently the decorator is not. Do do this, we can wrap this whole get method along with the decorator inside an instance method. That way we can get a `wait_gen` generator that is able to determine how long to wait for, depending on the Retry-After value.

snopoke: Yea, I was thinking of something like that. It might get ugly though. We'd also need to modify the `response.raise_for_status()` call so that we don't do double backoff.

I'm fine with this for now if you would rather try the other approach later.

SmittieC: Thoughts on this approach? https://github.com/dimagi/commcare-export/pull/225/commits/388a5077df9bada11af0ebf29f6ff28080d1ebe1

snopoke: Looks good to me

snopoke (approved):

snopoke (approved):

snopoke:
A unit test would be great

tests/test_commcare_hq_client.py
```diff
@@ -291,6 +291,26 @@ def test_message_log(self):
             [1, 2, 3]
         )
 
+    @patch("commcare_export.commcare_hq_client.CommCareHqClient.session")
```
SmittieC, line 294: Not sure what else we'd want to test?

snopoke (approved):

PR #230, opened 2024-02-02
Show detailed exception only when needed

mkangia:
https://dimagi.atlassian.net/browse/SC-3354

`raise for status` raises an exception in case of authentication failures which leads to a full stacktrace shown.
Seems reasonable to limit showing the full stacktrace only for verbose mode.

### DIFF
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -9,6 +9,7 @@ from __future__ import (
 
 import copy
 import logging
+import sys
 import time
 from collections import OrderedDict
 from math import ceil
@@ -151,7 +152,15 @@ class CommCareHqClient(object):
                 resource_url, params=params, auth=self.__auth, timeout=60
             )
             if self._should_raise_for_status(response):
-                response.raise_for_status()
+                try:
+                    response.raise_for_status()
+                except Exception as e:
+                    # for non-verbose output, skip the stacktrace
+                    if not logger.isEnabledFor(logging.DEBUG):
+                        logger.error(str(e))
+                        sys.exit()
+                    raise e
+
             return response
 
         response = _get(resource, params)
--- a/commcare_export/commcare_hq_client.py
+++ b/commcare_export/commcare_hq_client.py
@@ -159,7 +159,7 @@ class CommCareHqClient(object):
                     if not logger.isEnabledFor(logging.DEBUG):
                         if isinstance(e, requests.exceptions.HTTPError) and response.status_code == 401:
                             logger.error(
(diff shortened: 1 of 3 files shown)


zandre-eng:
I like that the stack trace is still accessible when in verbose mode, however there is a point from the Jira ticket that is still missing in the PR. Specifically, we need to add a more user friendly error message that can provide details on what the user can check to resolve the error. This could simply be the suggested error message as in the ticket:
> “401 Client Error: Unauthorized for URL [URL]. Please ensure that your CommCare HQ username and API Key are accurate, and verify that your account has the necessary permissions to access the DET tool.”

zandre-eng (commented):
I like that the stack trace is still accessible when in verbose mode, however there is a point from the Jira ticket that is still missing in the PR. Specifically, we need to add a more user friendly error message that can provide details on what the user can check to resolve the error. This could simply be the suggested error message as in the ticket:
> “401 Client Error: Unauthorized for URL [URL]. Please ensure that your CommCare HQ username and API Key are accurate, and verify that your account has the necessary permissions to access the DET tool.”

mkangia:
Thanks for the feedback @zandre-eng 

I totally missed that. I thought the request was to show specific error. I will update that now

zandre-eng (approved):
Looks good. One comment, although not blocking.

commcare_export/commcare_hq_client.py
```diff
@@ -157,7 +157,14 @@ def _get(resource, params=None):
                 except Exception as e:
                     # for non-verbose output, skip the stacktrace
                     if not logger.isEnabledFor(logging.DEBUG):
-                        logger.error(str(e))
+                        if isinstance(e, requests.exceptions.HTTPError) and response.status_code == 401:
+                            logger.error(
+                                f"#{e}. Please ensure that your CommCare HQ credentials are correct & valid for "
+                                f"the auth-mode passed. Also, Verify that your account has the necessary "
+                                f"permissions to access the DET tool."
```
kaapstorm: ("the DET tool" means "the data export tool tool".) But I'm wondering whether the user knows that Dimagi internally refers to this as "the DET". In help text this command refers to itself as "commcare-export" or "the commcare-export tool". I think this text might be more familiar to a user:
```suggestion
                                f"permissions to use commcare-export."
```

mkangia: Makes sense. Thank you

https://github.com/dimagi/commcare-export/pull/230/commits/5d99505143702c1bc320adec8635a134e01eeebb

kaapstorm (approved):

PR #232, opened 2024-02-05
Generate DET executable files

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-3351)

This PR broadens the DET horizons by introducing functionality to generate/compile executable files of the DET, each compatible with a different operating system. The supported operating systems are
1. Linux
2. Windows

Note that these two executable files are compiled automatically on every release publish and will be added as assets to the release once the workflow completes. Any person can then simply download the file(s) from there and use it (through the CLI) like you would as if you installed commcare-export,
> commcare-export < options >

The big push for this was the fact that some folks using the DET is less technically-inclined and having to manage a python codebase and it's packages poses a barrier to usage. Now users can simply download and run the file without thinking about any installation steps.

### DIFF
--- a/.github/workflows/release_actions.yml
+++ b/.github/workflows/release_actions.yml
@@ -1,0 +1,33 @@
+name: commcare-export release actions
+on:
+  release:
+    types: [published]
+
+jobs:
+  generate_release_assets:
+    name: Generate release assets
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Pull docker images for pyinstaller/wine
+        run: |
+          docker pull cdrx/pyinstaller-linux
+          docker pull cdrx/pyinstaller-windows
+
+      - name: Generate linux binary
+        run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/release_actions.yml
```diff
@@ -0,0 +1,33 @@
...
+          docker run -v "$(pwd):/src/" cdrx/pyinstaller-linux
+
+      - name: Generate windows exe
+        run: |
+          docker run -v "$(pwd):/src/" cdrx/pyinstaller-windows
+
+      - name: Upload release assets
+        uses: AButler/upload-release-assets@v3.0
+        with:
+          files: "./dist/linux/*;./dist/windows/*"
```
SmittieC, line 32: where would we find this? Or does it automatically attach the asset to the release?

SmittieC (approved):

.github/workflows/release_actions.yml
Charl1996: Yes, the files are automatically attached to the release as assets.

Charl1996:
Putting this back into draft as I have discovered something which will not make this work (yet).

Charl1996:
Will open new (working) PR

PR #233, opened 2024-02-12
Cs/sc 3373 det log file

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-3373)

The feature in this PR will become more important when [this ticket](https://dimagi.atlassian.net/browse/SC-3351) is implemented, i.e. when we're able to create executable files from the DET.

What this PR does is the following:
- Add a `Logger` class which allows stderr to log to same log file than other logs.
- Tunnel all `stderr` messages through to the logger so it will show up in the logs and not in the terminal, unless the user specifies a new flag, `no-logfile`, which will show all output in the terminal and not in the log file.

### DIFF
--- a/commcare_export/__init__.py
+++ b/commcare_export/__init__.py
@@ -1,1 +1,25 @@
+import sys
+import logging
 from .version import __version__
+
+logging.basicConfig(
+    filename="commcare_export.log",
+    format='%(asctime)s %(name)-12s %(levelname)-8s %(message)s',
+    filemode='w',
+)
+
+
+class Logger:
+    def __init__(self, logger, level):
+        self.logger = logger
+        self.level = level
+        self.linebuf = ''
+
+    def write(self, buf):
+        for line in buf.rstrip().splitlines():
+            self.logger.log(self.level, line.rstrip())
+
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+sys.stderr = Logger(logging.getLogger(), logging.ERROR)
--- a/commcare_export/__init__.py
+++ b/commcare_export/__init__.py
@@ -1,5 +1,8 @@
 import logging
+import os
 from .version import __version__
+
+repo_root = os.path.abspath(os.path.join(__file__, os.pardir, os.pardir))
 
 
 class Logger:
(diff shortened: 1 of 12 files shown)


kaapstorm (approved):

kaapstorm:
It's not obvious to me why tests are failing, but I think they are passing on master.

Charl1996:
Yeah, still need to look into it a little. I'm holding off this PR to make 100% it works as expected with the executable.

commcare_export/__init__.py
```diff
@@ -13,9 +16,26 @@ def write(self, buf):
             self.logger.log(self.level, line.rstrip())
 
 
+def logger_name_from_filepath(filepath):
```
kaapstorm, line 19: A test for this would be cool. ... Also, I suspect unusual things happen to filenames that start with "py", like "commcare_export/pyexample.py".

Charl1996: See [af99b49](https://github.com/dimagi/commcare-export/pull/233/commits/af99b494b5bea41d595612ca3f9a98c25b51d119)

kaapstorm (approved):
Nice! Thank you for obliging me.

commcare_export/__init__.py
```diff
@@ -21,7 +21,7 @@ def logger_name_from_filepath(filepath):
     return (
         relative_path.
         replace('/', '.').
-        replace('.py', '')
+        strip('.py')
```
kaapstorm, line 24: The `strip()` method strips any of the given characters from either side of a string.
```
>>> foo = 'pyppy.excel_query.py'
>>> foo.strip('.py')
'excel_quer'
>>> foo.rstrip('.py')
'pyppy.excel_quer'
```

So `rstrip()` is closer to what you want, but will still be too aggressive with `excel_query.py` because "excel_query" ends in a "y".

What you're really looking for is
```
>>> import re
>>> re.sub(r'\.py$', '', foo)
'pyppy.excel_query'
```

Charl1996: @kaapstorm 
Woah, thanks for illuminating this. I was definitely not aware of the full blast radius that `strip()` had. I always thought that it should match the string _exactly_ in order for the function to strip it...good reminder to double check those functions you thought you knew...

Charl1996: I added a new test with the example you gave...just to be sure...

kaapstorm (approved):

PR #235, opened 2024-02-14
Neaten up README.md a little

kaapstorm:
Addresses a few nits:
* Uses consistent spelling for "Python" and "CommCare HQ"
* Drops a broken link
* Fixes JSON and Python errors (except for one undefined variable that I couldn't figure out at a glance)
* Simplifies virtualenv instructions a little
* Applies recommended table formatting
* Updates GitHub workflow to run tests against current Python releases

This is a documentation only change.


.github/workflows/test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           - 5432:5432
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, '3.10', '3.11']
+        python-version: [3.8, 3.9, '3.10', 3.11, 3.12]
```
mkangia, line 34: nit: '3.10' instead of 3.10?

README.md
```diff
@@ -21,59 +21,58 @@ Once on a terminal window or command line, for simplicity, run commands from the
...
+Check which Python version is installed.
 
+This tool is tested with Python versions from 3.8 to 3.12.
 
 ```shell
 $ python --version
 $ python3 --version
 ```
-If python is installed, all of its available versions would be listed.
+If Python is installed, its version will be listed.
```
mkangia, line 31: nit:
version will be ~listed~ shown

README.md
```diff
@@ -21,59 +21,58 @@ Once on a terminal window or command line, for simplicity, run commands from the
...
 
-If python isn't installed, Install Python 3.8 from [this link](https://www.python.org/downloads/). 
+If Python isn't installed, [download and install](https://www.python.org/downloads/) the latest release.
```
mkangia, line 33: should we say install any version from 3.8 to 3.12?
If the latest version is above 3.12, It could lead to unexpected errors?

README.md
```diff
@@ -21,59 +21,58 @@ Once on a terminal window or command line, for simplicity, run commands from the
...
-to avoid conflicts with other python applications.
+to avoid conflicts with other Python applications.
 
 More about virtualenvs on https://docs.python.org/3/tutorial/venv.html
 
 Setup a virtual environment using:
 
 ```shell
-$ python3.8 -m venv .commcare-export-venv # update python version for the one installed
+$ python3 -m venv .venv
```
mkangia, line 45: Any reason to change that back to venv?
I had made that change earlier to make the name more specific and avoid conflicts with any other setup since that is the name suggested by virtualenv documentation by default. We use the same name for CommCareHQ, `venv`.

README.md
```diff
@@ -705,9 +705,9 @@ access to the corpora domain.
 
 These need to be set as environment variables as follows:
 
-```
-export HQ_USERNAME=<username>
-export HQ_API_KEY=<apikey>
+```shell
+$ export HQ_USERNAME=<username>
+$ export HQ_API_KEY=<apikey>
```
mkangia, line 711: nit:
Should not we generally avoid "$"?
Whenever I need to simply copy the command and run it, I need to remove that $ sign each time after pasting that into terminal.

README.md
```diff
@@ -356,20 +357,20 @@ query = Emit(
...
 
-print json.dumps(query.to_jvalue(), indent=2)
+print(json.dumps(query.to_jvalue(), indent=2))
 
 results = query.eval(BuiltInEnv() | CommCareHqEnv(api_client) | JsonPathEnv())
 
 if len(list(env.emitted_tables())) > 0:
-    # with writers.Excel2007TableWriter("excel-output.xlsx") as writer:
-    with writers.StreamingMarkdownTableWriter(sys.stdout) as writer:
+    # with Excel2007TableWriter("excel-output.xlsx") as writer:
```
mkangia, line 366: nit: could also remove that comment I believe.

README.md
```diff
@@ -448,9 +448,9 @@ referred to be name using `Ref`, and utilized via `Apply`.
 
 List of builtin functions:
 
-| Function                       | Description                                                                    | Example Usage                    |
-|--------------------------------|--------------------------------------------------------------------------------|----------------------------------|
-| `+, -, *, //, /, >, <, >=, <=` | Standard Math                                                                  |                                  |
+| Function                       | Description                                                                  | Example Usage                    |
+|--------------------------------|------------------------------------------------------------------------------|----------------------------------|
+| `+, -, *, //, /, >, <, >=, <=` | Standard Math                                                                |                                  |
```
mkangia, line 454: nit: Looks like this needs some more alignment correction.

.github/workflows/test.yml
```diff
@@ -47,8 +47,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - run: sudo apt-get install pandoc
-      - run: python setup.py sdist
       - run: pip install --upgrade pip
+      - run: pip install setuptools
```
mkangia, line 51: will this cause any trouble if running for version < 3.12?

README.md
```diff
@@ -71,6 +71,7 @@ $ venv
 Install CommCare Export via `pip`
 
 ```shell
+$ pip install setuptools
```
mkangia, line 74: nit: Should we make a comment that this is for version 3.12?

mkangia (commented):
Looks great! Thank for you this.

Left just minor queries.

.github/workflows/test.yml
kaapstorm: The quotes are required for 3.10 because without them YAML evaluates it as a float with the value 3.1. The others are fine.

README.md
kaapstorm: Personally, I like "venv". (Some people seem to like to hide the directory.) I'll change it.

README.md
kaapstorm: I added them in here to be consistent with the rest of the document (although I've just noticed I missed two others).

Generally, I have found it useful to indicate the prompt to make it clear what kind of prompt it is. e.g.
```
$ psql -U commcarehq

=# \l
    List of databases
...
=# \q

$ python3

>>> print('Hello world.')
Hello world.
```

I have found that often writers will leave out the prompt so consensus is not universal, but that more often than not, they include it. Certainly across Dimagi docs, we seem to include it most of the time.

.github/workflows/test.yml
mkangia: oh wow, can we keep all as strings then?

README.md
mkangia: I am happy with venv if you wish to keep that.

.github/workflows/test.yml
kaapstorm: No. In Python < 3.12, setuptools [is installed automatically](https://github.com/python/cpython/issues/95299), so installing it explicitly here has no effect.

(It is actually only required in Python 3.12 because we are installing commcare-export using setup.py in the next line. setuptools is not required when commcare-export is installed with `pip install -e .`.)

README.md
kaapstorm: I just tested this in Python 3.12. Turns out we don't need setuptools if we install using `pip install commcare-export`. I'll drop this line.

.github/workflows/test.yml
kaapstorm: f7bc3f5

README.md
kaapstorm: dff0651

README.md
kaapstorm: dff0651

README.md
kaapstorm: 84fe33b

README.md
kaapstorm: a2f0bf6

README.md
kaapstorm: e78dddb

README.md
kaapstorm: 3974ae6

README.md
```diff
@@ -42,23 +43,23 @@ More about virtualenvs on https://docs.python.org/3/tutorial/venv.html
 Setup a virtual environment using:
 
 ```shell
-$ python3 -m venv .venv
+$ python3 -m venv venv
```
mkangia, line 46: oh, we are not keeping the folder hidden anymore.

Non blocking but for someone who does not understand these things well, they would get really confused of two "venv"s, about which one is command and which one is a folder.

mkangia (approved):

README.md
kaapstorm: Oh! I completely misunderstood [your comment](https://github.com/dimagi/commcare-export/pull/235#discussion_r1488934391),

> Any reason to change that back to venv?
> I had made that change earlier to make the name more specific and avoid conflicts with any other setup since that is the name suggested by virtualenv documentation by default. We use the same name for CommCareHQ, `venv`.

I thought you were asking me to change `.venv` to `venv` in order to be consistent with its name in CommCare HQ. Sorry for my confusion.

> avoid conflicts

[virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/) puts all your virtual environments in one place, so it's important to give them unique names. But the Standard Library's [venv](https://docs.python.org/3/library/venv.html) is simpler, and doesn't have a special place for all your virtual environments. In my experience, most people just use a subdirectory in the repository's root named "venv", so there is no risk of conflicts. This convention is used in [GitHub's template `.gitignore` file for Python projects](https://github.com/github/gitignore/blob/main/Python.gitignore#L122-L129). I've just applied the same convention here.

mkangia: Okay.

PR #239, opened 2024-04-03
Improve DET errors

mkangia:
https://dimagi.atlassian.net/browse/SC-3407

Improving on some validations around expected output filename for the passed in output format.

1. Taking hints from how the [writer](https://github.com/dimagi/commcare-export/blob/a4fcf3fb19021c0a4cb33d58dd44a9df346c4d7a/commcare_export/cli.py#L339) is picked for a format, validated the extensions for csv, xls and xlsx. CSV files are actually stored in a zip file. Earlier using csv extension for csv format only gave a [warning](https://github.com/dimagi/commcare-export/blob/a4fcf3fb19021c0a4cb33d58dd44a9df346c4d7a/commcare_export/cli.py#L346-L352), we are now making it mandatory.
2. Additionally, for other non-sql output (json/markdown), validated that an extension is present in output filename, if used.

This will cause errors if users accidentally use the incorrect filenames for passed output format and let users know the fix for the same.

Along with this, a minor change to the error message related to authentication, to let user know that they should also ensure they have access to the project space.

Tested locally.

Also requested for QA here, https://dimagi.atlassian.net/browse/QA-6333

### DIFF
--- a/commcare_export/cli.py
+++ b/commcare_export/cli.py
@@ -186,6 +186,12 @@ def main(argv):
 
     args = parser.parse_args(argv)
 
+    if args.output_format and args.output:
+        errors = []
+        errors.extend(validate_output_filename(args.output_format, args.output))
+        if errors:
+            raise Exception(f"Could not proceed. Following issues were found: {', '.join(errors)}.")
+
     if not args.no_logfile:
         exe_dir = os.path.dirname(sys.executable)
         log_file = os.path.join(exe_dir, "commcare_export.log")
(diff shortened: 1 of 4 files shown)


zandre-eng (approved):
Looks good, nice tests!

mkangia:
Thanks for the review @zandre-eng 
Holding off on merge till QA is concluded, https://dimagi.atlassian.net/browse/QA-6333

mkangia:
QA passed.

PR #247, opened 2025-05-14
Drop undocumented "profile" arg

kaapstorm:
This is a small change that drops the "profile" command line argument.

It's undocumented, and it hasn't worked since we dropped support for Python 2, so we can be confident that it's not currently being used.

Also, some minor clean-ups:
* Changed a unit test.
* Updated the Python versions that the DET is tested with.


mkangia (approved):

ajeety4 (approved):
Nice :+1:

PR #269, opened 2026-04-10
Optimize SqlTableWriter 

kaapstorm:
Optimizes `SqlTableWriter`:

* Instead of running `make_table_compatible()` on every row, only runs on the first 10 (`SCHEMA_CHECK_ROWS` const) rows.
* Avoids increasing SQLAlchemy `MetaData` object growth.
* Upserts in batches of 1000 rows.
* Uses native bulk upserts when possible (PostgreSQL, MySQL) instead of one row at a time.
* Commits 1000 rows at a time instead of maintaining an open transaction for an entire export.


millerdev:
Suggestion: convert to DRAFT until you're done pushing to the branch

claude/specs/sql-writer-memory-optimization.md
```diff
@@ -0,0 +1,191 @@
...
+
+After row 10, a row may contain a value that is incompatible with the current
+column type, or a column that doesn't exist in the table. When a batch insert
+fails:
+
+1. Check whether the failure is due to a schema mismatch (missing column or
+   type incompatibility).
+2. If so, call `make_table_compatible()` for the failing row, then retry the
+   entire batch. Only do this **once per batch** -- if the retry also fails,
+   raise the original exception.
```
dannyroberts: If there are multiple rows in the same batch that have different schema mismatches, will the schema be updated for all rows in the batch that fail before retrying, or will it keep retrying until there are no more rows failing? The way it's written here it almost sounds like it'll try once, but if it fails on another row it'll just give up.

If the error we get back can only tell us about one row at a time, it almost seems better to just call `make_table_compatible()` for _all 1000_ rows and then retry once than to fix one, retry, and fail.

kaapstorm:
To make review easier, I'm going to split this into two PRs. I'll keep this one open, and chop it off after commit 7, [Clean up SQL writer memory optimization spec](https://github.com/dimagi/commcare-export/pull/269/changes/bf574fb04d943232fd8babd745d2d1ecadfd26f9).

commcare_export/writers.py
```diff
@@ -641,6 +641,24 @@ def bulk_upsert(self, table, batch):
...
+            sqlalchemy.exc.OperationalError,
+            sqlalchemy.exc.ProgrammingError,
+        ):
+            # Likely a schema mismatch; fix schema and retry once
+            for row_dict in batch:
+                table = self.make_table_compatible(
+                    table,
+                    row_dict,
+                    data_type_dict,
+                )
```
dannyroberts, line 682: Okay this answers my earlier question on the spec.

commcare_export/writers.py
```diff
@@ -619,11 +693,46 @@ def write_table(self, table_spec: TableSpec) -> None:
...
+                self.upsert(table, row_dict)
+            else:
+                assert table is not None  # So that mypy knows it's a Table
+                if not schema_check_complete:
+                    schema_check_complete = True
+                    self._commit()
```
dannyroberts: Will this work if there are exactly `SCHEMA_CHECK_ROWS` rows? Looking at the rows after the loop, I think so? But it's just a tad hard to follow.

commcare_export/writers.py
```diff
@@ -619,11 +693,46 @@ def write_table(self, table_spec: TableSpec) -> None:
...
+        else:
+            # All rows in schema-check phase; commit them
+            self._commit()
+
+        if not schema_check_complete:
+            logger.debug(
+                "Schema check complete for table '%s'. Final columns: %s",
+                table_name,
+                [c.name for c in table.columns],
+            )
```
dannyroberts: I wonder if we can do something nice with itertools that makes this code flow more naturally. Right now there's a lot of iteration accounting with each step doing conditional cleanup for the last step and/or lookahead for the next in a way that's a bit hard to follow.

I'm thinking something like this:

```python
if table_spec.rows:
    # ... create_table(...) ...

row_stream = iter(table_spec.rows)

for row in itertools.islice(row_stream, SCHEMA_CHECK_ROWS):
    table = self.make_table_compatible...
    self.upsert...

logger.debug("Schema check complete...

for batch in itertools.batched(row_stream, BATCH_SIZE):
    self._flush_batch(table, batch, ...
```

I think that's it—I think there'd then be no cleanup/commit/flush necessary afterwards.

dannyroberts (commented):
I had one substantive comment that I think will help a lot with the readability of one section, but otherwise looks good!

commcare_export/writers.py
kaapstorm: So much better! f7fba6e

commcare_export/writers.py
```diff
@@ -641,6 +641,24 @@ def bulk_upsert(self, table, batch):
...
                 self.upsert(table, row_dict)
 
+    def _flush_batch(self, table, batch, data_type_dict):
+        try:
+            self.bulk_upsert(table, batch)
+        except (
+            sqlalchemy.exc.CompileError,
+            sqlalchemy.exc.OperationalError,
+            sqlalchemy.exc.ProgrammingError,
+        ):
```
jingcheng16, line 675: Is this all possible exceptions when there is a incompatibility?

commcare_export/writers.py
```diff
@@ -673,11 +679,46 @@ def write_table(self, table_spec: TableSpec) -> None:
...
+                )
+                self.upsert(table, row_dict)
+            else:
+                assert table is not None  # So that mypy knows it's a Table
+                if not schema_check_complete:
+                    schema_check_complete = True
+                    self._commit()
+                    logger.debug(
+                        "Schema check complete for table '%s'. Final columns: %s",
+                        table_name,
```
jingcheng16: Can we be explicit that this is just a schema check for the first `SCHEMA_CHECK_ROWS` rows? This might be helpful for people who didn't read the code and just read the log.

commcare_export/writers.py
```diff
@@ -615,6 +615,20 @@ def _commit(self):
...
+        # SQLAlchemy requires all dicts in `batch` to have the same keys
+        # for `insert(table).values(batch)`. We need to drop the columns
+        # whose values are always `None` to reproduce the behavior of
+        # `SqlTableWriter.insert()`. `batch_keys` are the columns where
+        # _any_ row has a value set.
+        batch_keys = {
+            k for row_dict in batch
+            for k, v in row_dict.items()
+            if v is not None
+        }
```
jingcheng16: I find this set comprehension hard to read...maybe it is just me...

jingcheng16 (approved):
No blocking comments.

commcare_export/writers.py
jingcheng16: AI comment: 
The old upsert() (writers.py:585) does:                                                                                                                           
 ```                                                                                                                                                                                                              
    row_dict = {col: val for col, val in row_dict.items() if val is not None}                                                                                                                                  
    # then UPDATE ... SET (only the non-None cols)    
```                                                                                                                                                         
So an existing row's columns that aren't in row_dict are preserved.                                                                                                                                        
                                                                                                                                                                                                               
                                                                                                                                                                                                               
The new `bulk_upsert()` builds `batch_keys` as the union of non-None keys across the whole batch, then in PG iterates `stmt.excluded` and in MySQL iterates `table.columns` to construct the SET clause. Worth verifying whether columns that exist in the table but were dropped from `batch_keys` (because every row in that batch had None for them) end up getting clobbered to `NULL` on update. Old code wouldn't have touched them; new code might.

Thoughts? Sounds like a real concern

File: commcare_export/writers.py [hunk 8]
```diff
@@ -605,25 +608,100 @@ def upsert(self, table, row_dict):
...
+
+        rows = (dict(zip(headings, row)) for row in table_spec.rows)
+        first_row = next(rows, None)
+        if first_row is None:
+            return
+        row_stream = itertools.chain([first_row], rows)
+
+        table = self.get_table(table_name)
+        if table is None:
+            table = self.create_table(table_name, first_row, data_type_dict)
```
dannyroberts, line 699: Nit: Would the following work? I think the only difference would be that `self.make_table_compatible` would not be called on the first row, but that might be okay (just based on the names of the functions) if create_table is already taking into account `first_row`'s schema? I guess technically we'd also need `SCHEMA_CHECK_ROWS - 1` below. What I like about this is that then we're literally doing one full clean sweep through the data, no lookahead or backtracking, so there's no data flow to keep in your head. Process first row. Process next N rows. Process rest.

Not a big deal at all; the way you have it is not much harder to understand.

```suggestion
        row_stream = (dict(zip(headings, row)) for row in table_spec.rows)
        first_row = next(row_stream, None)

        if first_row is None:
            return

        table = self.get_table(table_name)
        if table is None:
            table = self.create_table(table_name, first_row, data_type_dict)
            self.upsert(table, first_row)
```

dannyroberts (approved):
Very nice! The itertools refactor is so very much easier to follow! Left one small thought on the same topic, but totally up to you—looks good as is.

commcare_export/writers.py
kaapstorm: 804873e

commcare_export/writers.py
kaapstorm: dd54c0f

commcare_export/writers.py
kaapstorm: d3f6566

commcare_export/writers.py
kaapstorm: Nicely spotted! ca0221e

commcare_export/writers.py
kaapstorm: > I think the only difference would be that self.make_table_compatible would not be called on the first row.

Yes, and neither would `self.upsert` ([on the next line](https://github.com/dimagi/commcare-export/blob/ca0221e591231a754ed1eaa57d926e5473b5ed82/commcare_export/writers.py#L703)), so the first row would not end up in the database.

jingcheng16 (approved):

PR #272, opened 2026-04-28
Upgrade SQLAlchemy from 1.4 to 2.0

kaapstorm:
This branch upgrades SQLAlchemy from 1.4 to 2.0. This is required by the changes in PR https://github.com/dimagi/commcare-export/pull/269 for the transaction management introduced in SQLAlchemy 2.0.

Each commit is small, or covers a small change across many files. Automated changes by ruff are in their own commits.


commcare_export/checkpoint.py
```diff
@@ -47,7 +47,7 @@ def get_pagination_mode(self):
         if not self.pagination_mode:
             return PaginationMode.date_modified
 
-        return PaginationMode[self.pagination_mode]
+        return PaginationMode[str(self.pagination_mode)]
```
dannyroberts: It feels wrong to call `str` when we already expect `self.pagination_mode` to be a string, just to fix a type checking error. The only other value it could take is `None`, and in that situation coercing to `'None'` wouldn't be the right "fix", right?

Would something like this fix the type checker warning?
```suggestion
        if self.pagination_mode is None:
            return PaginationMode.date_modified
        else:
            return PaginationMode[self.pagination_mode]
```

kaapstorm: c430eef + 93a3333 : Use SQLAlchemy 2.0 typed column definitions, and move typing into its own module to make the model definition more readable.

jingcheng16 (dismissed):
I know nothing about SQLAlchemy, so I just reviewed and make sure the code changes follow the commit messages. Can approve again when conflicts are resolved.

jingcheng16 (approved):

PR #273, opened 2026-07-16
Two miscellaneous changes

kaapstorm:
Apropos nothing, this PR makes two small changes:

* Drops a sentence from `CLAUDE.md` that suggests formatting.
* Removes an accidentally-committed empty zip file. It also adds it to `.gitignore` because it is the default filename of outputted data and should be ignored.


.gitignore
```diff
@@ -57,3 +57,6 @@ uv.lock
 
 # Build artifacts
 *.whl
+
+# Default output filename
+reports.zip
```
millerdev: Can this be created anywhere in the repo, or only at the root? If the latter


```suggestion
/reports.zip
```

millerdev (approved):

.gitignore
kaapstorm: https://github.com/dimagi/commcare-export/pull/273/commits/5b861caae4984eb00d9f6d4a629fa33c6e9f6561

.gitignore
```diff
@@ -1,62 +1,13 @@
...
+/dist/
 
 # uv
-.uv/
-uv.lock
-
-# Build artifacts
-*.whl
+/.venv/
+/uv.lock
```
millerdev, line 9: I see it didn't change here, but `uv.lock` should not be ignored.

millerdev (approved):

.gitignore
kaapstorm: I agree. I'm intending to unignore it when I bump the release.

